PR #12, opened 2017-11-16
Use pip-tools for requirements management

### LINKED ISSUE #5: Use pip-tools for requirements management

jonc125:
Fixes #5.

Also added some basic dev instructions to the readme.


MichaelClerx:
Is there a way to tell `pip` that we want the Python3 versions of all of these? Or to specify a minimum Python version?

jonc125:
You can set a min python version in setup.py. Or test on both 2 & 3 with travis.

PR #20, opened 2018-01-18
Implement remaining CellML 2.0 tags

### LINKED ISSUE #9: Expand the range of supported MathML
We should probably support the [CellML subset](https://www.cellml.org/specifications/cellml_1.1/#sec_math_cellml_subset) eventually (see also in [CellML 2](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit#heading=h.zegia3mrify8)). Not urgent though - just the elements in the models we have currently is sufficient to get started :)

tamuri:
- Fixes #9
- Supports MathML tags in CellML 2.0 spec.
- MathML tags mapping directly to Sympy classes are handled by single handler. Reduces boilerplate code considerably.
- Some more tests added for simple operators

### DIFF
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -318,7 +318,7 @@ def get_nary_relation_callback(sympy_relation):
     """
     Wraps the Sympy binary relation to handle n-ary MathML relations
 
-    :param sympy_relation: handle to binary Sympy relation (Eq, Le, Lt, Ge, Gt)
+    :param sympy_relation: handle for binary Sympy relation (Eq, Le, Lt, Ge, Gt)
     :return: callback used by the apply_handler to handle n-ary relations
     """
     def _wrapper_relational(*expressions):
@@ -326,7 +326,7 @@ def get_nary_relation_callback(sympy_relation):
         if len(expressions) > 2:
             # Convert to multiple Sympy binary relations bundled in an 'And' boolean
             equalities = []
-            for first, second in zip(expressions, expressions[1:]):
+            for first, second in zip(expressions[:-1], expressions[1:]):
                 equalities.append(sympy_relation(first, second))
             return sympy.And(*equalities)
         return sympy_relation(*expressions)
@@ -342,7 +342,7 @@ def simple_operator_handler(node):
 
     handler = getattr(sympy, SIMPLE_MATHML_TO_SYMPY_NAMES[tag_name])
 
-    # Some MathML relations allow chaining but Synpy relations are binary
+    # Some MathML relations allow chaining but Sympy relations are binary operations
     if tag_name in MATHML_NARY_RELATIONS:
         return get_nary_relation_callback(handler)
 
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -3,6 +3,7 @@ Parses Content MathML and returns equivalent SymPy expressions
 
 Content Markup specification: https://www.w3.org/TR/MathML2/chapter4.html
 """
+import logging
 from xml.dom import Node, minidom
 
 import sympy
@@ -45,12 +46,13 @@ def transpile(xml_node):
             # (see cn_handler for an example), show a message
             text = child_node.data.strip()
             if text:
-                print('Hit text node with text "' + text + '"')
+                logging.warning('Unhandled text node in <%s>: "%s"', child_node.tagName, text)
         elif child_node.nodeType == child_node.ELEMENT_NODE:
             # Call the appropriate MathML handler function for this tag
             tag_name = child_node.tagName
             if tag_name in HANDLERS:
                 sympy_expressions.append(HANDLERS[tag_name](child_node))
+                logging.debug('Transpiled node %s ⟶ %s', child_node.toxml(), sympy_expressions[-1])
             else:
                 # MathML handler function not found for this tag!
                 raise NotImplementedError('No handler for element <%s>' % tag_name)
@@ -115,6 +117,9 @@ def apply_handler(node):
     https://www.w3.org/TR/MathML2/chapter4.html#contm.apply
     """
     result = transpile(node)
+
+    logging.debug('Result of <apply>:\n\t%s\t⟶\t%s', node.toxml(), result)
+
     if len(result) > 1:
         expression = result[0](*(result[1:]))
     else:
@@ -170,13 +175,13 @@ def minus_handler(node):
     * Negation (-a) is equivalent to sympy.Mul(sympy.S.NegativeOne, a)
     * Subtraction (a - b) is equivalent to sympy.Add(a, sympy.Mul(sympy.S.NegativeOne, b))
     """
-    def wrapped_minus(left_operand, right_operand=None):
+    def _wrapped_minus(left_operand, right_operand=None):
         if right_operand is None:
             # unary arithmetic operator => negation
             return -left_operand
         # otherwise, binary arithmetic operator => subtraction
         return left_operand - right_operand
-    return wrapped_minus
+    return _wrapped_minus
 
 
 def divide_handler(node):
@@ -187,9 +192,9 @@ def divide_handler(node):
 
     Equivalent to sympy.Mul(a, sympy.Pow(b, sympy.S.NegativeOne))
     """
-    def wrapped_divide(dividend, divisor):
+    def _wrapped_divide(dividend, divisor):
         return dividend / divisor
-    return wrapped_divide
+    return _wrapped_divide
 
 
 def power_handler(node):
@@ -198,9 +203,9 @@ def power_handler(node):
     binary arithmetic operator
     equivalent to sympy.Pow(a, b)
     """
-    def wrapped_power(base, exponent):
+    def _wrapped_power(base, exponent):
         return base ** exponent
-    return wrapped_power
+    return _wrapped_power
 
 
 def root_handler(node):
@@ -214,12 +219,12 @@ def root_handler(node):
     root element. Thus, square roots correspond to the case where degree contains the value 2, cube
     roots correspond to 3, and so on. If no degree is present, a default value of 2 is used.
     """
-    def wrapped_root(first_argument, second_argument=None):
+    def _wrapped_root(first_argument, second_argument=None):
         # if no <degree> given, it's sqrt
         if second_argument is None:
             return sympy.root(first_argument, 2)
         return sympy.root(second_argument, first_argument)
-    return wrapped_root
+    return _wrapped_root
 
 
 def degree_handler(node):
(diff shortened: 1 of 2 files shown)


cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -236,72 +217,32 @@ def root_handler():
...
-    n-ary operator
-    """
-    return sympy.And
+    Nasty:
+    The root element is used to construct roots. The kind of root to be taken is specified by a
+    degree element, which should be given as the second child of the apply element enclosing the
+    root element. Thus, square roots correspond to the case where degree contains the value 2, cube
+    roots correspond to 3, and so on. If no degree is present, a default value of 2 is used.
+    """
+    def __wrapped_root(first_argument, second_argument=None):
```
jonc125: I'm not sure there's any benefit to using `__` here - it's not a class method that needs to be private, just a local variable that gets returned, at which point the name doesn't matter.

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -236,72 +217,32 @@ def root_handler():
...
+    https://www.w3.org/TR/MathML2/chapter4.html#contm.degree
+    Meaning of <degree> depends on context! We implement it for order of <bvar> in <diff> and
+    the kind of root in <root>
     """
-    return sympy.Or
+    result = transpile(node)
+    if len(result) != 1:
+        raise ValueError('Expected single value in <degree> tag.'
+                         'Got: ' + node.toxml())
+    return transpile(node)[0]
```
jonc125: `return result[0]`

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -324,112 +274,137 @@ def bvar_handler(node):
...
-def arccos_handler():
+def simple_operator_handler(tag_name):
     """
-    https://www.w3.org/TR/MathML2/chapter4.html#contm.trig
-    unary trigonometric operator
-    """
-    return sympy.acos
-
+    This function handles simple MathML <tagName> to sympy.Class operators, where no unique handling
+    of tag children etc. is required. The mappings of tagName to sympy class names is in
```
jonc125: Think you're missing the end of the sentence here!

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -324,112 +274,137 @@ def bvar_handler(node):
...
-            'piece': piece_handler,
-            'piecewise': piecewise_handler,
-            'plus': plus_handler,
-            'power': power_handler,
-            'rem': rem_handler,
-            'root': root_handler,
-            'tanh': tanh_handler,
-            'times': times_handler}
+# These MathML tags map directly to Sympy class and don't require any extra handling
+SIMPLE_MATHML_TO_SYMPY_NAMES = {
```
jonc125: I've not checked all the Sympy names against their documentation... assuming they're correct!

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -48,15 +48,19 @@ def transpile(xml_node):
...
             # Call the appropriate MathML handler function for this tag
-            name = child_node.tagName
-            if name in HANDLERS:
+            tag_name = child_node.tagName
+            if tag_name in HANDLERS:
                 # If this tag element itself has children
                 if child_node.childNodes:
                     # We want to pass the node to the handler, and it will deal with children
-                    sympy_expressions.append(HANDLERS[name](child_node))
+                    sympy_expressions.append(HANDLERS[tag_name](child_node))
```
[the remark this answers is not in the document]
tamuri: Great, yes!

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -91,8 +94,26 @@ def cn_handler(node):
...
-    TODO: 'type' attribute?
     """
+
+    # If this number is using scientific notation
+    if 'type' in node.attributes:
+        if node.attributes['type'].value == 'e-notation':
+            # A real number may also be presented in scientific notation. Such numbers have two
+            # parts (a mantissa and an exponent) separated by sep. The first part is a real number,
+            # while the second part is an integer exponent indicating a power of the base.
+            # For example, 12.3<sep/>5 represents 12.3 times 105. The default presentation of
```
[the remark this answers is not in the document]
tamuri: Ctrl+C Ctrl+V error! 😦

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -91,8 +94,26 @@ def cn_handler(node):
...
+            # this example is 12.3e5.
+            if len(node.childNodes) == 3 and node.childNodes[1].tagName == 'sep':
+                mantissa = float(node.childNodes[0].data.strip())
+                exponent = int(node.childNodes[2].data.strip())
+                return sympy.Float('%fe%d' % (mantissa, exponent))
```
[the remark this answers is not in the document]
tamuri: True, will fix. I’ll leave exponent wrapped in int()

cellmlmanip/mathml2sympy/transpiler.py
tamuri: That’s the bit you’re meant to check!! 😉

jonc125: OK!

tamuri: Before you drown in MathML & Sympy documentation…jk!

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -48,18 +48,12 @@ def transpile(xml_node):
...
-            if name in HANDLERS:
-                # If this tag element itself has children
-                if child_node.childNodes:
-                    # We want to pass the node to the handler, and it will deal with children
-                    sympy_expressions.append(HANDLERS[name](child_node))
-                else:
-                    # This tag has no children
-                    sympy_expressions.append(HANDLERS[name]())
+            tag_name = child_node.tagName
+            if tag_name in HANDLERS:
```
MichaelClerx: This could just be 

```
try:
    sympy_expressions.append(HANDLERS[tag_name](child_node))
catch KeyError:
    raise NotImplemented...
```

(More "pythonic", but also avoids an extra dictionary lookup)

cellmlmanip/mathml2sympy/transpiler.py
MichaelClerx: Two underscores is reserved for python internals anyway, so should be a single underscore if anything

tests/test_mathml2sympy.py
```diff
@@ -154,6 +179,37 @@ def test_diff_eq(self):
...
+
+    def test_min(self):
+        self.assert_equal('<apply><min/><ci>a</ci><ci>b</ci></apply>',
+                          [sympy.Min(sympy.Symbol('a'), sympy.Symbol('b'))])
+
+    def test_max(self):
+        self.assert_equal('<apply><max/><ci>a</ci><ci>b</ci></apply>',
+                          [sympy.Max(sympy.Symbol('a'), sympy.Symbol('b'))])
+
+    def test_trig(self):
```
MichaelClerx: Are we not doing the full trig set cellml supports? So (arccos hyperbolicus etc. :p)

tests/test_mathml2sympy.py
```diff
@@ -118,12 +125,30 @@ def test_log(self):
...
+        self.assert_equal('<apply>'
+                          '<diff/>'
+                          '<bvar><ci>time</ci><degree><cn>2</cn></degree></bvar>'
+                          '<ci>V</ci>'
+                          '</apply>',
+                          [sympy.Derivative(V(time), time, 2)])
+
     def test_piecewise(self):
         x = sympy.Symbol('x')
         self.assert_equal('<piecewise>'
```
MichaelClerx: Should we add a longer one with more options than one? Maybe conflicting ones too to check if they're implemented the same.

For example:
if x > 0: y = 10
if x > 1: y = 20
if x > 2: y = 30
else: y= 0

This should return 10 for x = 3, because that's the first matched option

tests/test_mathml2sympy.py
```diff
@@ -58,6 +57,10 @@ def test_eq(self):
         self.assert_equal('<apply><eq/><ci>a</ci><ci>b</ci></apply>',
                           [sympy.Eq(sympy.Symbol('a'), sympy.Symbol('b'))])
```
MichaelClerx: Maybe add an eq with 3 arguments. MathML parses (0 == 0 == 1) as false, (because 0 isn't 1), whereas C will see it as `((0 == 0) == 1) = (1 == 1) = 1`. Python does it like MathML, so guessing sympy does too?

tests/test_mathml2sympy.py
tamuri: They're there but only tested a few.

tests/test_mathml2sympy.py
tamuri: Thanks, that's a case I didn't consider. Looks like Sympy relations all handle 2 arguments out of the box but MathML relations are all n-ary. I'll need to implement this.

tamuri: Related: https://github.com/sympy/sympy/issues/6059

tests/test_mathml2sympy.py
tamuri: Looks like Sympy does that.

```
>>> import sympy
>>> x = sympy.Symbol('x')
>>> expr = sympy.Piecewise((10, x > 0), (20, x > 1), (30, x > 2), (0, True))
>>> expr.subs({x: 3})
10
```

cellmlmanip/mathml2sympy/transpiler.py
tamuri: Cool - single underscore silences warnings about docstrings in my IDE!

cellmlmanip/mathml2sympy/transpiler.py
jonc125: That has the danger that if the handler exists but raises `KeyError` for some reason when called, you'll get a misleading error message. `in` is pretty quick for a dict so performance is unlikely to be an issue. If it turns out to be, we could always do something like
```python
handler = HANDLERS.get(tag_name, None)
if handler:
    sympy_expressions.append(handler(child_node))
else:
    raise ...
```

tests/test_mathml2sympy.py
jonc125: Good test to add for the model evaluation part.

tamuri:
> I take it And is n-ary then?

Yes, thankfully!

tests/test_mathml2sympy.py
tamuri: Fixed in latest commit

tamuri:
Are we happy to merge this PR?

jonc125:
Yep

Issue #21, opened 2018-02-21
Model manipulation


jonc125:
As of Sept 16 2019 I (@skeating ) have broken up the remaining issues here into smaller issues:
#80 #81 #82 #83 #84 
Any unchecked item below reference the issue that now also contains it.

Initial thoughts, to be expanded...

* [x] Parsing of simple mathematical systems (cf #22) - Issues copied to #80
* [x] Parsing of realistic models - Issues copied to #82
* [x] Analysing systems of equations - Issues copied to #81
* [x] Interface needed for model manipulation  - Issues copied to #83
* [x] Interface needed for code generation - Issues copied to #84

Tasks:
- [x] @jonc125 to create initial sample CellML files - see #22

jonc125:
The easiest way to test the parsing parts of this is probably to simulate the parsed models (cf #2) which potentially also requires the code generation aspects. Need to discuss (tomorrow!) how much overlap/integration we want between these tools. See also https://github.com/ModellingWebLab/project_issues/issues/12 and discussions about [Melodee](https://github.com/LLNL/melodee).

jonc125:
@tamuri can you tick off in the description the items now done?

tamuri:
ticked off a few more items

MichaelClerx:
Shall we remove the sub-items here for every section that's now got its own overview ticket?

jonc125:
We could remove the detailed text from here, but I'd keep the links to the individual issues.

MauriceHendrix:
By the way the way the code for chaste codegen does it at the moment is without replacing variables.
say V is in Volt then:
- I find V and multiply by conversion factor
- use subs on other equations to substitute vs for 1/factor * V

It's probably less easy/safe but it does get round any issues with ordering of the variables

PR #22, opened 2018-02-21
Sample CellML files for testing [WIP]

jonc125:
For #21.


tamuri:
`<xsd:choice>` means only one of the elements, but that one element can occur multiple times with `maxOccurs="unbounded"`. The CellML DTD also defines the same choice but has the occurence as you describe: `<!ELEMENT connection (map_components | map_variables+)>`

tamuri:
> I checked the files with pycml and it says they're OK, which I trust much more :)

Thanks & confirmed, validates with the Relax NG schema from https://chaste.comlab.ox.ac.uk/cellml/

PR #25, opened 2018-06-21
feature/cellml parsing

### LINKED ISSUE #7: Import maths from a CellML file
Not trying to do any CellML validation here (that's down to libCellML) just getting the maths from components and connecting up variables.

- [ ] Extract maths from each component and convert separately
- [ ] Ensure that variables in different components with the same name don't accidentally get equated by Sympy's symbol cache
- [ ] Process connections and add `sympy.Eq` accordingly
- [ ] Import units definitions (requires #8)

tamuri:
This is a work in progress (PR for review & feedback).

This branch introduces the CellML model parser and allows some manipulation of equations, handling of units & unit conversion. The is no "main" method as yet - I've been developing off the tests (test_parser.py).

Will fix #7.

### DIFF
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -18,11 +18,25 @@ class Transpiler(object):
     """Transpiler class handles conversion of MathmL to Sympy exprerssions"""
 
     def __init__(self, dummify: bool = False, symbol_prefix: str = None, symbol_lookup=dict()) -> None:
+        # we create symbols as necessary, as they occur in equations
+        self.error_on_unknown_symbol = False
+
+        # create dummy symbols for variables rather than typical sympy symbols
+        self.dummify: bool = dummify
+
+        # use to store information about dummified numbers (number & units)
         self.metadata: Dict = dict()
 
-        self.dummy_symbol_cache = symbol_lookup
-        self.dummify: bool = dummify
+        # prefix all symbols with given string
         self.symbol_prefix = symbol_prefix
+
+        # for a given variable, get it's symbol from this dictionary
+        self.dummy_symbol_cache = symbol_lookup
+
+        # if a symbol lookup dictionary was supplied
+        if symbol_lookup:
+            # unknown variables in equations are an error
+            self.error_on_unknown_symbol = True
 
         # Mapping MathML tag element names (keys) to appropriate handler for SymPy output (values)
         # These tags require explicit handling because they have children or context etc.
@@ -120,8 +134,13 @@ class Transpiler(object):
         if self.symbol_prefix:
             identifier = self.symbol_prefix + identifier
         if self.dummify:
-            # Return a dummified version of this symbol, picking up from cache
-            return self.dummy_symbol_cache.setdefault(identifier, sympy.Dummy(identifier))
+            if self.error_on_unknown_symbol:
+                assert identifier in self.dummy_symbol_cache, \
+                    '%s not found in symbol dict' % identifier
+                return self.dummy_symbol_cache[identifier]
+            else:
+                # Return a dummified version of this symbol, picking up from cache
+                return self.dummy_symbol_cache.setdefault(identifier, sympy.Dummy(identifier))
         return sympy.Symbol(identifier)
 
     def _cn_handler(self, node):
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -1,5 +1,4 @@
-"""
-Parses Content MathML and returns equivalent SymPy expressions
+"""Parses Content MathML and returns equivalent SymPy expressions
 
 Content Markup specification: https://www.w3.org/TR/MathML2/chapter4.html
 """
@@ -67,15 +66,13 @@ class Transpiler(object):
             self.handlers[tag_name] = self._simple_operator_handler
 
     def parse_string(self, xml_string) -> List[sympy.Expr]:
-        """
-        Reads MathML content from a string and returns equivalent SymPy expressions
+        """Reads MathML content from a string and returns equivalent SymPy expressions
         """
         dom = minidom.parseString(xml_string)
         return self.parse_dom(dom.childNodes[0])
 
     def parse_dom(self, math_dom_element) -> List[sympy.Expr]:
-        """
-        Accepts a <math> node of DOM structure and returns equivalent SymPy expressions.
+        """Accepts a <math> node of DOM structure and returns equivalent SymPy expressions.
         Note: math_dom_element must point the <math> XmlNode, not the root XmlDocument
 
         :param math_dom_element: <math> XmlNode object of a MathML DOM structure
@@ -84,8 +81,7 @@ class Transpiler(object):
         return self.transpile(math_dom_element)
 
     def transpile(self, xml_node):
-        """
-        Descends the given MathML element node and calls the corresponding handler for child
+        """Descends the given MathML element node and calls the corresponding handler for child
         elements. Returns the SymPy expression of node
         :param xml_node: a DOM element of parsed MathML
         :return: a list of SymPy expressions
@@ -120,8 +116,7 @@ class Transpiler(object):
     # MATHML ELEMENT HANDLERS ######################################################################
 
     def _math_handler(self, node):
-        """
-        Descend XML node <math>...</math>
+        """Descend XML node <math>...</math>
         """
         result = self.transpile(node)
         return result
@@ -129,8 +124,7 @@ class Transpiler(object):
     # TOKEN ELEMENTS ###############################################################################
 
     def _ci_handler(self, node):
-        """
-        MathML:  https://www.w3.org/TR/MathML2/chapter4.html#contm.ci
+        """MathML:  https://www.w3.org/TR/MathML2/chapter4.html#contm.ci
         SymPy: http://docs.sympy.org/latest/modules/core.html#id17
         """
         identifier = node.childNodes[0].data.strip()
@@ -147,8 +141,7 @@ class Transpiler(object):
         return sympy.Symbol(identifier)
 
     def _cn_handler(self, node):
-        """
-        MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.cn
+        """MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.cn
         SymPy: http://docs.sympy.org/latest/modules/core.html#number
         """
 
@@ -185,8 +178,7 @@ class Transpiler(object):
     # BASIC CONTENT ELEMENTS #######################################################################
 
     def _apply_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.apply
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.apply
         """
         result = self.transpile(node)
 
@@ -199,8 +191,7 @@ class Transpiler(object):
         return expression
 
     def _piecewise_handler(self, node):
-        """
-        MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
+        """MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
         SymPy: http://docs.sympy.org/latest/modules/functions/elementary.html#piecewise
 
         constructor, zero or more <piece>, zero or one <otherwise>
@@ -209,8 +200,7 @@ class Transpiler(object):
         return sympy.Piecewise(*result)
 
     def _piece_handler(self, node):
-        """
-        MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
+        """MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
         Returns a 2-tuple defining an expression and condition
         <piece> element contains exactly two children
         """
@@ -220,8 +210,7 @@ class Transpiler(object):
         return result[0], result[1]
 
     def _otherwise_handler(self, node):
-        """
-        MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
+        """MathML: https://www.w3.org/TR/MathML2/chapter4.html#contm.piecewise
         Returns a 2-tuple defining an expression and condition
         """
         result = self.transpile(node)
@@ -232,8 +221,7 @@ class Transpiler(object):
     # ARITHMETIC, ALGEBRA AND LOGIC ################################################################
 
     def _minus_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.minus
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.minus
         unary arithmetic operator OR binary arithmetic operator
 
         From http://docs.sympy.org/latest/tutorial/manipulation.html:
@@ -252,8 +240,7 @@ class Transpiler(object):
         return _wrapped_minus
 
     def _divide_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.divide
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.divide
         binary arithmetic operator
         There is no class in SymPy for division. Rather, division is represented by a power of -1.
 
@@ -264,8 +251,7 @@ class Transpiler(object):
         return _wrapped_divide
 
     def _power_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.power
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.power
         binary arithmetic operator
         equivalent to sympy.Pow(a, b)
         """
@@ -274,8 +260,7 @@ class Transpiler(object):
         return _wrapped_power
 
     def _root_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.root
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.root
         operator taking qualifiers
 
         Nasty:
@@ -292,8 +277,7 @@ class Transpiler(object):
         return _wrapped_root
 
     def _degree_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.degree
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.degree
         Meaning of <degree> depends on context! We implement it for order of <bvar> in <diff> and
         the kind of root in <root>
         """
@@ -306,8 +290,7 @@ class Transpiler(object):
     # CALCULUS AND VECTOR CALCULUS #################################################################
 
     def _diff_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.diff
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.diff
         operator taking qualifiers
         TODO: clean up dummification if possible
         """
@@ -341,8 +324,7 @@ class Transpiler(object):
         return _wrapped_diff
 
     def _bvar_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.bvar
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.bvar
         NASTY: bvar element depends on the context it is being used
         In a derivative, it indicates the variable with respect to which a function is being
         differentiated.
@@ -361,8 +343,7 @@ class Transpiler(object):
     # ELEMENTARY CLASSICAL FUNCTIONS ###############################################################
 
     def _log_handler(self, node):
-        """
-        https://www.w3.org/TR/MathML2/chapter4.html#contm.log
+        """https://www.w3.org/TR/MathML2/chapter4.html#contm.log
         operator taking qualifiers or a unary calculus operator
         """
         def _wrapped_log(first_element, second_element=None):
@@ -375,8 +356,7 @@ class Transpiler(object):
         return _wrapped_log
 
     def _logbase_handler(self, node):
-        """
-        Qualifier for <log>
+        """Qualifier for <log>
 
         The log function accepts only the logbase schema. If present, the logbase schema denotes the
         base with respect to which the logarithm is being taken. Otherwise, the log is assumed to be
@@ -389,8 +369,7 @@ class Transpiler(object):
         return self.transpile(node)[0]
 
     def _get_nary_relation_callback(self, sympy_relation):
-        """
-        Wraps the Sympy binary relation to handle n-ary MathML relations
+        """Wraps the Sympy binary relation to handle n-ary MathML relations
 
         :param sympy_relation: handle for binary Sympy relation (Eq, Le, Lt, Ge, Gt)
         :return: callback used by the apply_handler to handle n-ary relations
@@ -407,8 +386,7 @@ class Transpiler(object):
         return _wrapper_relational
 
     def _simple_operator_handler(self, node):
-        """
-        This function handles simple MathML <tagName> to sympy.Class operators, where no unique
+        """This function handles simple MathML <tagName> to sympy.Class operators, where no unique
         handling of tag children etc. is required.
         """
         tag_name = node.tagName
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -17,7 +17,10 @@ logger.setLevel(logging.INFO)
 class Transpiler(object):
     """Transpiler class handles conversion of MathmL to Sympy exprerssions"""
 
-    def __init__(self, dummify: bool = False, symbol_prefix: str = None, symbol_lookup=dict()) -> None:
+    def __init__(self,
+                 dummify: bool = False,
+                 symbol_prefix: str = None,
+                 symbol_lookup=dict()) -> None:
         # we create symbols as necessary, as they occur in equations
         self.error_on_unknown_symbol = False
 
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -17,9 +17,10 @@ logger.setLevel(logging.INFO)
 class Transpiler(object):
     """Transpiler class handles conversion of MathmL to Sympy exprerssions"""
 
-    def __init__(self, dummify: bool = False, symbol_prefix: str = None) -> None:
+    def __init__(self, dummify: bool = False, symbol_prefix: str = None, symbol_lookup=dict()) -> None:
         self.metadata: Dict = dict()
-        self.dummy_symbol_cache = dict()
+
+        self.dummy_symbol_cache = symbol_lookup
         self.dummify: bool = dummify
         self.symbol_prefix = symbol_prefix
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -222,17 +222,30 @@ class Variable(object):
     """Represents a <variable> in a CellML model. A component may contain any number of <variable>
     elements, which define variables that may be mathematically related in the equation blocks
     contained in the component. """
-    def __init__(self, *, name, units, symbol: sympy.Symbol=None, initial_value=None,
+    def __init__(self, *, name, units, dummy: sympy.Symbol=None, initial_value=None,
                  public_interface=None, private_interface=None, **kwargs):
         self.name = name
         self.units = units
         self.initial_value = initial_value
         self.public_interface = public_interface
         self.private_interface = private_interface
-        self.symbol = symbol
+        self.dummy = dummy
         self.cmeta_id = kwargs.get('cmeta:id', None)
         self._component_name = kwargs.get('_component_name', None)
         self._original_name = kwargs.get('_original_name', None)
+
+    def __str__(self) -> str:
+        return '%s(%s)' % (
+            type(self).__name__,
+            ', '.join('%s=%s' % item for item in vars(self).items() if item[1])
+        )
+
+
+class NumberWrapper(object):
+    def __init__(self, dummy, units, number):
+        self.dummy = dummy
+        self.units = units
+        self.number = number
 
     def __str__(self) -> str:
         return '%s(%s)' % (
@@ -256,6 +269,8 @@ class Model(object):
         self.dummy_info: Dict[Dict] = defaultdict(dict)
         self.graph: nx.DiGraph = None
         self.variables_x: Dict[str, Variable] = OrderedDict()
+        self.equations_x: List[sympy.Eq] = list()
+        self.numbers_x: Dict[sympy.Dummy, NumberWrapper] = dict()
 
     def __str__(self):
         """Pretty-print each of the components in this model"""
@@ -270,6 +285,18 @@ class Model(object):
         else:
             self.units.add_custom_unit(units_name, unit_attributes)
 
+    def add_equation(self, equation: sympy.Eq):
+        assert isinstance(equation, sympy.Eq)
+        self.equations_x.append(equation)
+
+    def add_number(self, dummy: sympy.Dummy, attributes: Dict):
+        assert 'sympy.Number' in attributes
+        assert isinstance(dummy, sympy.Dummy)
+        assert dummy not in self.numbers_x
+        self.numbers_x[dummy] = NumberWrapper(dummy,
+                                              attributes['cellml:units'],
+                                              attributes['sympy.Number'])
+
     def add_component(self, component: Component):
         """Adds name to list of <component>s in the <model>
         """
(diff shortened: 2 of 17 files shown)


tests/test_parser.py
```diff
@@ -0,0 +1,149 @@
...
+        assert len(model.components) == 17  # grep -c '<component ' test_simple_odes.cellml
+
+    def test_equations_count(self, model):
+        equation_count = 0
+        for component in model.components.values():
+            equation_count += len(component.equations)
+        assert equation_count == 16  # determined by hand
+
+    def test_variable_find(self, model):
+        assert model.find_variable({'cmeta:id': 'time'}) == [{'cmeta:id': 'time',
```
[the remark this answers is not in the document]
tamuri: We can change if we think we're going to add code/logic into there. I'll leave as dict for now.

tests/test_parser.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Custom units defined in CellML example
+        assert units.convert_to(model.units.get_quantity('per_ms'), 1/units.millisecond) == 1/units.millisecond
+        assert units.convert_to(model.units.get_quantity('usec'), units.microsecond) == units.microsecond
+        assert units.convert_to(
+            model.units.get_quantity('mM_per_ms'),
+            [units.mole, units.liter, units.millisecond]) == (units.mole / 1000) / (units.liter * units.millisecond)
+
+    def test_add_units_to_equations(self, model):
+        # This is an irreversible operation # TODO: don't mutate?
```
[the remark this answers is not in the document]
tamuri: Yes, this inserts the units into the equations (not only for numbers, all symbols). Equations to convert are added when we make the connections.

tests/test_parser.py
```diff
@@ -0,0 +1,149 @@
...
+        assert QuantityStore.is_equal(lhs_units, new_rhs_units)
+
+        # Try fixing all units on the RHS so that they match the LHS
+        for _, component in model.components.items():
+            for index, equation in enumerate(component.equations):
+                lhs_units = QuantityStore.summarise_units(equation.lhs)
+                rhs_units = QuantityStore.summarise_units(equation.rhs)
+                if not QuantityStore.is_equal(lhs_units, rhs_units):
+                    new_rhs = units.convert_to(equation.rhs, lhs_units)
+                    # Create a new equality with the converted RHS and replace original
```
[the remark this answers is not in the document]
tamuri: Not as yet - need to take this logic (once we're happy conversions are working) and put it into the model handling.

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        if not self.equations:
+            return
+
+        # Get all unique symbols in the equations
+        dummy_symbols = set()
+        for equation in self.equations:
+            for free_symbol in equation.free_symbols:
+                dummy_symbols.add(free_symbol)
+            # Collect bound variables for derivatives (not returned by .free_symbols)
+            for derivative in equation.atoms(sympy.Derivative):
```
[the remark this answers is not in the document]
tamuri: Yes, returns all the atoms for all Derivatives wherever they are in the equation.

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+                dummy_symbols.add(derivative.variables[0])  # TODO: variables for degree > 1
```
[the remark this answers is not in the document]
tamuri: Transpiler has code to parse mathml higher-derivatives, which would be nice to leave. We could throw an error outside that somewhere?

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        if expr_unit_info:
+            expr_unit_info = self.model.units.get_quantity(expr_unit_info)
+            unit_info = {expr: expr_unit_info}
+            logging.debug('Found unit for "%s" ⟶ "%s"', expr, expr_unit_info)
+        else:
+            unit_info = {}
+
+        # Traverse down the arguments of this expression, and collect their unit information
+        for args in expr.args:
+            unit_info: Dict = {**unit_info, **self._collect_units(args)}
```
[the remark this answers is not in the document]
tamuri: Agreed. Not sure why I did it this way... 😕

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+
+        variable_with_dummy = self.find_variable({'sympy.Dummy': expr})
+        if variable_with_dummy:
+            return variable_with_dummy[0]['units']
+
+        for dummy_number, number_info in self.numbers.items():
+            if dummy_number == expr:
+                return number_info['cellml:units']
+
+        # Note unit information about this expression in this component. It might be referring to
```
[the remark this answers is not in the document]
tamuri: It's to handle the time variable in the environment component, which doesn't have any `<math>` element. If that's the only case, perhaps it's better to be explicit about what we're doing here.

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+    def add_rdf(self, rdf: str):
+        """ Takes RDF string and stores it in an RDFlib.Graph for the model. Can be called
+        repeatedly.
+        """
+        self.rdf.parse(StringIO(rdf))
+
+    def _is_free_variable(self, variables: Dict):
+        """Checks whether a variable gets its value for another component. There are two
+        possibilities. Either is has:
+            (i) public_interface="out" or
```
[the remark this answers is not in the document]
tamuri: Does a `public=out private=in` variable allow it to have a connection with another variable from within the same component?

cellmlmanip/parser.py
```diff
@@ -0,0 +1,139 @@
...
+            # Add the component instance to the model
+            self.model.add_component(component)
+
+    def __add_maths(self, component: Component, component_element: etree.Element):
+        """ <model> <component> <math> </component> </model> """
+        # NOTE: Only looking for one <math> element
+        math_element = component_element.find(Parser.with_ns(XmlNs.MATHML, u'math'))
+        if math_element is not None:
+            transpiler = mathml2sympy.Transpiler(dummify=True, symbol_prefix=component.name + '__')
+            # TODO: check whether element can be passed directly without .tostring()
```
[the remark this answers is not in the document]
tamuri: It's not drop-in, unfortunately, need to do some massaging to get it to work.

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        :param search_dict:
+        :return: a dictionary of key: value pairs
+        """
+        # a list to collect all matched variables
+        matches = []
+
+        # for each defined variable in the component
+        for variable in self.variables.values():
+            # Every search (key, value) pair needs to be in the variable's attributes
+            matched = True
```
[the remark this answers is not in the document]
tamuri: Added check (above)

cellmlmanip/model.py
```diff
@@ -137,11 +141,20 @@ def _find_unit_info(self, expr):
...
+        cmeta:id etc. Search criteria are described as key-value pairs in a dictionary. Those
+        model variables that fulfil _all_ key-value pairs in the search_dict are returned.
+
+        :param search_dict: a dictionary providing the search criteria
+        :return: a list of variables in the model matching the search criteria
         """
+
+        # throw error if empty search criteria given
+        if search_dict is None or not search_dict:
+            raise RuntimeError("Search criteria cannot be empty.")
```
jonc125: `ValueError` might be more appropriate here; I think that's typically used for an invalid argument to a function.

cellmlmanip/model.py
```diff
@@ -186,6 +199,9 @@ def add_unit(self, units_elements: dict):
     def add_component(self, component: Component):
         """Adds name to list of <component>s in the <model>
         """
+        if component.name in self.components:
+            raise RuntimeError("%s already exists. Check CellML." % component.name)
```
jonc125: Similarly here; generally Python style says try to avoid the more generic error classes. I think `ValueError` is the best fit from https://docs.python.org/3/library/exceptions.html

cellmlmanip/model.py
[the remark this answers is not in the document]
jonc125: Not if I understand the question correctly. So say you have components A-E where A is the parent of B & C, and B is the parent of D & E, with variables as follows:
- D.v public=out
- E.v public=in connects to D.v
- B.v private=in public=out connects to D.v and A.v and C.v
- C.v public=in connects to B.v
- A.v private=in connects to B.v

In other words, when you have `public=out private=in` you don't need to export the variable privately (and indeed can't) since your child components will be siblings of the original exporter and can get the variable directly from there.

cellmlmanip/model.py
```diff
@@ -362,61 +378,135 @@ class QuantityStore(object):
...
+        'kelvin',
+        'kilogram',
+        'meter',
+        'metre',  # in self.UNIT_ALIASES
+        'mole',
+        'second',
+
+        # Derived SI units
+        'becquerel',  # see __add_custom_units()
+        'celsius',  # TODO: not defined by Sympy
```
jonc125: I'd instead add a comment saying this is removed in CellML 2 and won't be supported by us :)

tests/test_parser.py
```diff
@@ -142,8 +144,11 @@ def test_unit_extraction(self):
...
         assert QuantityStore.summarise_units(eq) == (millivolts / units.milliseconds)
 
-    def _test_print(self, model):
-        # show equations
-        for _, component in model.components.items():
-            for equation in component.equations:
-                print(equation)
+    def test_print(self, model):
+        from os import environ
+        if "CMLM_TEST_PRINT" in environ:
```
jonc125: I meant using pytest's actual skip functionality :)

tamuri: Aah, OK!! 👍

cellmlmanip/model.py
tamuri: Gladly!

cellmlmanip/model.py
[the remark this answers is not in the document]
tamuri: I'm handling group encapsulated relationships now, and have updated the __connect() method accordingly. Could you check to see if I've got the logic right?

cellmlmanip/model.py
```diff
@@ -243,20 +274,19 @@ def make_connections(self):
...
-                    # If it doesn't have a dummy symbol
-                    if 'sympy.Dummy' not in var_attr:
-                        # This variable was not used in any equations - create a new dummy symbol
-                        var_attr['sympy.Dummy'] = var_attr['assignment'] = sympy.Dummy(
-                            component.name + SYMPY_SYMBOL_DELIMITER + var_attr['name'])
-                    else:
-                        # The variable is used in an equation & we use the same symbol
-                        var_attr['assignment'] = var_attr['sympy.Dummy']
+
+        # For each CellML <variable< in the model
```
MichaelClerx: typo!

cellmlmanip/model.py
```diff
@@ -436,18 +470,37 @@ def __connect(self, connection):
...
+        def __are_siblings(name_a, name_b):
+            return self.components[name_a].parent == self.components[name_b].parent
+
+        def __parent_of(parent_name, child_name):
+            return parent_name == self.components[child_name].parent
+
+        def __has_interface(variable_a, key_a, value_a, variable_b, key_b, value_b):
+            return (key_a, value_a) in variable_a.items() and (key_b, value_b) in variable_b.items()
+
+        # if one component is encapsulated by the other
```
MichaelClerx: The relevant lines from the spec are

> The set of all components immediately encapsulated by the current component is the encapsulated set.

> Other components encapsulated by the same parent make up the sibling set.

> The interface exposed to the parent component and components in the sibling set is defined by the public_interface attribute. The private_interface attribute defines the interface exposed to components in the encapsulated set. Each interface has three possible values: "in", "out", and "none", where "none" indicates the absence of an interface.

might be good to quote those in a comment!

cellmlmanip/model.py
```diff
@@ -436,18 +470,37 @@ def __connect(self, connection):
...
+        if __parent_of(cmp_name_1, cmp_name_2) or __parent_of(cmp_name_2, cmp_name_1):
+            # they can be connected with (i) public=out to private=in (ii) private=out to public=in
```
MichaelClerx: I think this is correct

cellmlmanip/model.py
```diff
@@ -436,18 +470,37 @@ def __connect(self, connection):
...
+            if __has_interface(*(var_att_1, pri, 'in'), *(var_att_2, pub, 'out')):
+                return self.__connect_with_direction(cmp_name_2, var_att_2, cmp_name_1, var_att_1)
+            elif __has_interface(*(var_att_1, pri, 'out'), *(var_att_2, pub, 'in')):
+                return self.__connect_with_direction(cmp_name_1, var_att_1, cmp_name_2, var_att_2)
+
+        # if components are siblings (in group or top-level group) or one encapsulates the other
```
MichaelClerx: Again, think this is right

cellmlmanip/model.py
jonc125: I'm not sure here. I think you want to separate out the cases depending on which is the parent - the lines below only work if comp1 is the parent.

cellmlmanip/model.py
jonc125: Again, I don't think so! public-public only works for siblings, not encapsulation.

cellmlmanip/model.py
```diff
@@ -0,0 +1,813 @@
...
+            elif __has_interface(*(var_att_2, pub, 'out'), *(var_att_1, pub, 'in')):
+                return self.__connect_with_direction(cmp_name_2, var_att_2, cmp_name_1, var_att_1)
+
+        raise ValueError("Cannot determine the source & target for connection %s" % str(connection))
+
+    def __connect_with_direction(self, source_component, source_variable,
+                                 target_component, target_variable):
+        """Given the source and target component and variable, create a connection by assigning
+        the symbol from the source to the target. If units are not the same, it will add an equation
+        to the component reflecting the relationship. If a symbol has not been assigned to the
```
jonc125: To which component?

cellmlmanip/model.py
tamuri: Will fix (and the next one)

MichaelClerx: Maybe stick in a test that this code fails first

jonc125: I'm guessing just swapping the components around in `<map_components component_1="circle_y" component_2="circle_y_implementation"/>` would make this fail.

cellmlmanip/model.py
jonc125:
```diff
--- i/tests/cellml_files/err_connect_to_hidden_component.cellml
+++ w/tests/cellml_files/err_connect_to_hidden_component.cellml
@@ -6,21 +6,30 @@


    <component name="parent">
+      <variable name="time2" public_interface="in" units="second"/>
    </component>

    <component name="child">
       <variable name="time" public_interface="out" units="second"/>
+      <variable name="time2" public_interface="out" units="second"/>
    </component>

    <component name="other">
       <variable name="time" public_interface="in" units="second"/>
    </component>

+   <!-- This isn't allowed as child is hidden from other -->
    <connection>
       <map_components component_1="child" component_2="other"/>
       <map_variables variable_1="time" variable_2="time"/>
    </connection>

+   <!-- This isn't allowed as the private_interface within parent is none -->
+   <connection>
+      <map_components component_1="child" component_2="parent"/>
+      <map_variables variable_1="time2" variable_2="time2"/>
+   </connection>
+
    <group>
       <relationship_ref relationship="encapsulation"/>
       <component_ref component="parent">
```

tamuri: Re. your second connection above, isn't that allowed according to spec: 

> Variables with a public_interface attribute value of "in" must be mapped to a single variable in a component _in the encapsulated_ or sibling sets with a public_interface attribute value of "out" or to a single variable in the parent of the current component with a private_interface attribute value of "out".

jonc125: Hmm, that contradicts other parts of the spec I think! Will check when I get a chance...

tamuri: If I understand correctly, what you're saying is that  (given a valid in/out pair) siblings are connected on the `public_interface` and parent/child are connected on `public`-`private` interfaces.

jonc125: From https://www.cellml.org/specifications/cellml_1.0/#sec_structure_spec_rules_map_variables_element :
>  A variable with a public_interface attribute value of "in" must be mapped to a single variable owned by a component in the sibling set, provided the target variable has a public_interface attribute value of "out", or to a single variable owned by the parent component, provided the target variable has a private_interface attribute value of "out".
>  A variable with a public_interface attribute value of "out" may be mapped to variables owned by components in the sibling set, provided the target variables have public_interface attribute values of "in". It may also be mapped to variables owned by the parent component, provided the target variables have private_interface attribute values of "in".
> A variable with a private_interface attribute value of "in" may be mapped to a single variable owned by a component in the encapsulated set, provided the target variable has a public_interface attribute value of "out".
> A variable with a private_interface attribute value of "out" may be mapped to variables owned by components in the encapsulated set, provided the target variables have public_interface attribute values of "in".

cellmlmanip/parser.py
```diff
@@ -0,0 +1,195 @@
...
+        # handle the child elements of <model>
+        self.__add_units(model_xml)
+        self.__add_components(model_xml)
+        self.__add_relationships(model_xml)
+        self.__add_rdf(model_xml)
+        self.__add_connection(model_xml)
+
+        return self.model
+
+    def __add_rdf(self, element: etree.Element):
```
MichaelClerx: Why 2 leading underscores instead of 1? Normally this is reserved for python built-ins. (Also, 1 underscore already indicates hidden, why is that not enough?)

cellmlmanip/parser.py
```diff
@@ -0,0 +1,195 @@
...
+        """The main method that reads the XML file and extract the relevant parts of CellML model
+        definition. Parser class should have been instantiated with the filepath.
+
+        :return: a Model class holding CellML model definition, reading for manipulation
+        """
+        tree = etree.parse(self.filepath)
+
+        # <model> root node - initialise the model object
+        model_xml = tree.getroot()
+        self.model = Model(model_xml.get(Parser.with_ns(XmlNs.CMETA, u'id')))
```
MichaelClerx: I don't think we need `u'string'` anywhere; unicode literals are the default in Python 3 aren't they?

cellmlmanip/model.py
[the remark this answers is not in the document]
tamuri: Added assertion

cellmlmanip/parser.py
```diff
@@ -0,0 +1,139 @@
...
+            self.__add_variables(component, component_element)
+            self.__add_maths(component, component_element)
+            self.__add_rdf(component_element)
+
+            # Add the component instance to the model
+            self.model.add_component(component)
+
+    def __add_maths(self, component: Component, component_element: etree.Element):
+        """ <model> <component> <math> </component> </model> """
+        # NOTE: Only looking for one <math> element
```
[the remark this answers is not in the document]
MichaelClerx: https://github.com/MichaelClerx/cellml-validation/blob/master/models_1_0/valid/3.4.2.1.component_with_maths.cellml

cellmlmanip/parser.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Rename key for cmeta_id (remove namespace from attribute)
+            cmeta_id_attribute = Parser.with_ns(XmlNs.CMETA, 'id')
+            if cmeta_id_attribute in attributes:
+                attributes['cmeta:id'] = attributes.pop(cmeta_id_attribute)
+
+            component.variables[attributes['name']] = attributes
+
+            # Add any RDF for this <variable>
+            self.__add_rdf(variable_element)
```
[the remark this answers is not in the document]
tamuri: Now using `iter` which descends into subtree (`findall` only picks up immediate children)

cellmlmanip/parser.py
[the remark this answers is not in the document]
tamuri: Thanks - done!

cellmlmanip/parser.py
tamuri: You're right, I'll remove them

tamuri:
As discussed, I'm able to parse and check units for all models in FunctionalCuration except the following:

cellmlmanip/parser.py
tamuri: I've switched to use single underscore

MichaelClerx:
> 'grandi_pasqualini_bers_2010_ss_endo.cellml', # exponent of equation is a dummy, set by equation elsewhere

What does that mean exactly? Is the CellML invalid?

tamuri:
I think the CellML is valid but the expressions need to be evaluated to correctly determine the units e.g. in [this model](https://models.cellml.org/e/96/grandi_pasqualini_bers_2010.cellml/cellml_math) `Qpow` is assigned using an expression and then is itself exponent in other expressions. I think I need to eval the equations to determine units of rhs in those expressions.

cellmlmanip/model.py
```diff
@@ -0,0 +1,574 @@
...
+                # everywhere the target variable is used, replace with source variable
+                for index, equation in enumerate(self.equations):
+                    self.equations[index] = equation.xreplace(
+                        {target_variable.dummy: source_variable.assigned_to}
+                    )
+            else:
+                # Requires a conversion, so we add an equation assigning the target dummy variable
+                # to the source variable
+                self.equations.append(
+                    # TODO: do unit conversion here?
```
tamuri: See #46

tamuri:
Hide private members of Model

cellmlmanip/units.py
```diff
@@ -0,0 +1,432 @@
...
+                    out = self.ureg.Quantity(float(metadata.initial_value), metadata.units)
+                else:
+                    # otherwise, keep the symbol
+                    out = self.ureg.Quantity(expr, metadata.units)
+            return out
+        elif expr.is_Integer:
+            out = int(expr) * self.ureg.dimensionless
+            return out
+        elif expr.is_Rational:
+            # NOTE: can't send back Rational(1,2) * u.dimensionless
```
tamuri: Note: 
```
In [10]: sp.Rational(1,2) * u.dimensionless
Out[10]: 0.500000000000000

In [11]: u.Quantity(sp.Rational(1,2), u.dimensionless)
Out[11]: <Quantity(1/2, 'dimensionless')>
```

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -1,355 +1,404 @@
...
-    result = transpile(node)
-    if len(result) == 1:
-        # Bound variable without specifying degree
+                raise NotImplementedError('Unimplemented type attribute for <cn>: '
+                                          + node.attributes['type'].value)
+        else:
+            number = float(node.childNodes[0].data.strip())
+            number = sympy.Number(number)
+
+        if self.dummify:
```
MichaelClerx: Do we use both options in the code? (With and without self.dummify) ? Otherwise maybe this option can go

cellmlmanip/model.py
```diff
@@ -0,0 +1,574 @@
...
+                target_variable.assigned_to = target_variable.dummy
+            logger.debug('Updated target: %s', target_variable)
+            return True
+        # The source variable has not been assigned a symbol, so we can't make this connection
+        logger.info('The source variable has not been assigned to a symbol '
+                    '(i.e. expecting a connection): %s ⟶ %s',
+                    target_variable.name, source_variable.name)
+        return False
+
+    def add_rdf(self, rdf: str):
```
MichaelClerx: Should these be alphabetic? Or have we not decided on that @jonc125 ?

cellmlmanip/units.py
```diff
@@ -0,0 +1,451 @@
...
+    'zepto': 1e-21,
+    'atto':  1e-18,
+    'femto': 1e-15,
+    'pico':  1e-12,
+    'nano':  1e-9,
+    'micro': 1e-6,
+    'milli': 1e-3,
+    'centi': 1e-2,
+    'deci':  1e-1,
+    'deca':  1e+1,
```
MichaelClerx: @jonc125 Are you happy to have this in as a silent extra option? I think we'll be allowing deca and deka in CellML 2.0 (also nobody uses it)

setup.py
```diff
@@ -15,5 +15,19 @@
     author_email='',
     url='https://github.com/ModellingWebLab/cellmlmanip',
     license=license_,
-    find_packages=find_packages(exclude=('tests', 'docs'))
+    packages=find_packages(exclude=('tests', 'docs')),
+    package_data={'cellmlmanip': ['cellml_units.txt']},
+    include_package_date=True,
```
MichaelClerx: date --> data

setup.py
```diff
@@ -15,5 +15,19 @@
     author_email='',
     url='https://github.com/ModellingWebLab/cellmlmanip',
     license=license_,
-    find_packages=find_packages(exclude=('tests', 'docs'))
+    packages=find_packages(exclude=('tests', 'docs')),
+    package_data={'cellmlmanip': ['cellml_units.txt']},
```
MichaelClerx: Do we need this _and_ the manifest file? Maybe manifest file wasn't working because of typo below?

cellmlmanip/model.py
jonc125: What do you mean?

MichaelClerx: A lot of people do the methods in a class alphabetically.

jonc125: Oh, I see. I'm not strict on that, provided the order makes some sort of sense. I'll tend to search when I'm looking for something anyway! So having related methods adjacent (where you're likely to need to look at both when reading/changing the code) tends to be what I go for.

MichaelClerx: Cool! That's what I used to do but it turns out to be difficult to agree upon between people, and moving stuff around messes up your git history :D
(By which I mean, if you change _and move_ a method in a single commit, it looks like delete/insert, so git can't point to the modified lines anymore)

jonc125: Yeah, don't do that :)

First/lead author of the code gets to decide. Anyone else doesn't get to move methods except with agreement, and in a separate commit.

cellmlmanip/units.py
tamuri: Pint can't handle magnitude of sympy.Rational - leaving as is

cellmlmanip/parser.py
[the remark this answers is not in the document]
tamuri: Created issue #51 for this

tests/test_parser.py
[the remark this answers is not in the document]
tamuri: This has been refactored

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        for symbol in dummy_symbols:
+            # If this symbol is a number
+            if symbol in metadata and 'sympy.Number' in metadata[symbol]:
+                # save the number metadata (units etc.) and skip variable
+                self.numbers[symbol] = metadata[symbol]
+                continue
+
+            # if the symbol is one that's defined as a <variable> in the component
+            variable_name = symbol.name.split('__')[1]
+            if symbol.name.split('__')[1] in self.variables:
```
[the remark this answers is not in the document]
tamuri: This doesn't apply in the refactored code

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+                    unit_info[key] = self.numbers[key]['sympy.Number'] * value
+                # Otherwise, straightforward insertion of unit
+                else:
+                    unit_info[key] = key * value
+
+            # Do the substitutions (WARNING: irreversible!)
+            self.equations[index] = equation.subs(unit_info)
+
+    def _collect_units(self, expr):
+        # Find the unit information (if any) associated with this expression
```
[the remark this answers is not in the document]
tamuri: removed

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+                self.numbers[symbol] = metadata[symbol]
+                continue
+
+            # if the symbol is one that's defined as a <variable> in the component
+            variable_name = symbol.name.split('__')[1]
+            if symbol.name.split('__')[1] in self.variables:
+                self.variables[variable_name]['sympy.Dummy'] = symbol
+
+    def add_units_to_equations(self):
+        """Inserts unit attributes associated with symbols into equations
```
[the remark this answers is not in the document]
tamuri: removed

cellmlmanip/model.py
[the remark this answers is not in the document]
tamuri: no longer applicable

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        if variable_with_assigned:
+            return variable_with_assigned[0]['units']
+
+        # No information found about this
+        return None
+
+    def find_variable(self, search_dict):
+        """Finds variables in this model that match the given key: value pairs
+
+        :param search_dict:
```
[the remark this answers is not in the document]
tamuri: added note about metadummy

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        """
+        self.name: str = name
+        self.units: 'QuantityStore' = None
+        self.components: Dict[str, Component] = {}
+        self.connections: List[Tuple] = []
+        self.rdf: rdflib.Graph = rdflib.Graph()
+
+    def __str__(self):
+        """Pretty-print each of the components in this model"""
+        return '\n'.join([str(v) for k, v in self.components.items()])
```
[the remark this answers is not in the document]
tamuri: no longer applicable

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        """
+
+        # At this stage we are going to assign variables to new sympy Dummy placeholders
+        # This may change in the future once we're confident that everything is working e.g.
+        # manipulate the equations directly
+
+        # First, we assign new sympy.Dummy variables for those CellML <variable>s that are only
+        # exposed to other components i.e. do not have an "in" value on public/private_interface
+        # For each component in the model
+        for _, component in self.components.items():
```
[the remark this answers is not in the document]
tamuri: n/a

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        elif (('public_interface', 'out') in var_att_2.items() or
+              ('private_interface', 'out') in var_att_2.items()) and (
+                  ('public_interface', 'in') in var_att_1.items() or
+                  ('private_interface', 'in') in var_att_1.items()):
+            return self.__connect_with_direction(cmp_name_2, var_att_2, cmp_name_1, var_att_1)
+        raise RuntimeError("Cannot determine the source & target for connection %s" % connection)
+
+    def __connect_with_direction(self, source_component, source_variable,
+                                 target_component, target_variable):
+        """Given the source and target component and variable, create a connection by either
```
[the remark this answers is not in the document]
tamuri: n/a

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+    """Holds sympy.physics.unit.Quantity objects for the model. Can be initialised with <units>
+    definitions from the CellML <model>. The get_quantity() methods will find the right
+    quantity to return, given the name of the unit (from the internal store, Sympy built-in, or
+    constructed anew)
+    """
+
+    # Aliases for units to their Sympy equivalent
+    aliases = {
+        'litre': 'liter',
+        'litres': 'liters',
```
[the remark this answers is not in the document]
tamuri: removed

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        exec_('from sympy.physics.units import *', self.sympify_context)
+
+    def get_quantity(self, unit_name):
+        """Given the name of the unit, this will either (i) return a Quantity from the internal
+        store as its already been resolved previously (ii) return the Quantity from Sympy if it a
+        built-in name or (iii) construct and return a new Quantity object using the
+        <units><unit></units> definition in the CellML <model>
+        """
+        # This name is aliased to some other unit name (e.g. british -> american spelling)
+        if unit_name in QuantityStore.aliases:
```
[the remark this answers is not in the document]
tamuri: n/a

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+            unit_name = QuantityStore.aliases[unit_name]
+
+        # If we've already sourced the quantity for this name
+        if unit_name in self.store:
+            return self.store[unit_name]
+
+        # If this unit name is one of the built-in Sympy quantities
```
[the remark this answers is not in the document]
tamuri: n/a

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+        """
+        # If this expression is a Quantity itself
+        if isinstance(expr, units.Quantity):
+            # Return it as it is
+            return expr
+        # Don't descend into Derivative expression (there are no units within!)
+        if isinstance(expr, sympy.Derivative):
+            return None
+        # Units are always part of a Multiplicative expression
+        # TODO: check if units are always multiplicative!
```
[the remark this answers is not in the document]
tamuri: added to test_units.py

cellmlmanip/model.py
```diff
@@ -0,0 +1,508 @@
...
+                self.sympify_context[unit_element['units']] = unit_as_quantity
+
+            # Construct a string representing the expression and dimensions for this <unit>
+            expr = str(unit_as_quantity.name)
+            dimension = str(unit_as_quantity.args[1].args[0])
+
+            if 'prefix' in unit_element:
+                expr = '%s * %s' % (unit_element['prefix'], expr)
+
+            expr = '(%s)' % expr
```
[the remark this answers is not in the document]
tamuri: n/a

cellmlmanip/model.py
tamuri: this look ok, much clearer after the refactoring

cellmlmanip/mathml2sympy/transpiler.py
tamuri: It's used in unit tests for mathml2sympy (outside the context of cellml models)

setup.py
tamuri: not sure. some discussion about needing both, one for source distributions and the other for binary distributions

tamuri:
@jonc - if you're happy, I'll merge

jonc125:
Go for it!

Issue #31, opened 2018-11-16
Cellmlmanip is incompatible with newer versions of sympy


MichaelClerx:
It's currently pinned to 1.1.1 (Jul 27, 2017) but fails with 1.2 (Jul 9, 2018) and 1.3 (Sep 14, 2018), both of which are pip installable stable releases. 1.3 adds an additional deprecation warning

jonc125:
Is the new API better?

jonc125:
Hmm, that doesn't look particularly promising!

MichaelClerx:
It does seem to say they've only deprecated stuff, so maybe this isn't the cause of the issue we're seeing?

MichaelClerx:
@tamuri is this fixed in the new branch?

tamuri:
Haven't upgraded as yet. I'll work on it next week, I think it'll break some of the expression printing, so be prepared!

MichaelClerx:
ping!

tamuri:
I know there's discussion about pinning packages at all but I've updated the packages (Sympy required some changes to cellmlmanip)

https://github.com/ModellingWebLab/cellmlmanip/commit/da64d7f3e9c0b54e6eca44511fda7cf6a661e9f5 (feature/update-packages)

MichaelClerx:
Thanks!

Issue #30, opened 2018-11-16
setup.py is missing dependencies


MichaelClerx:
This means my [plan to install it via git into fccodegen](https://github.com/ModellingWebLab/fccodegen/blob/66c2046cb9b6db5c3ef3d95d72bb32fed953dcee/dev-requirements/base.in#L1) (until pip can do it) doesn't work.

Can we add the dependencies to setup.py? Preferably unpinned? @jonc125 @tamuri ?

jonc125:
Sounds good!

MichaelClerx:
Seems to be working now!

PR #32, opened 2018-11-16
Setup changes

MichaelClerx:
See #30 

Has a pinned version of sympy due to #31 

Also note this is a merge into cellml-parsing, so we can keep that as the branch for fccodegen :D

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
     - "3.6"
 install:
-    - pip install -e .[test]
+    - pip install .[test]
 script:
     - pytest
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,8 +2,7 @@ language: python
 python:
     - "3.6"
 install:
-    - pip install -r requirements/test.txt
-    - pip install -e .
+    - pip install -e .[test]
```
jonc125: Do the tests still pass without the `-e`? They should (and we should remove it).

MichaelClerx: They didn't! But fixed that now :-)

jonc125: And this is why Travis shouldn't use `-e` :)

MichaelClerx:
Any comments @tamuri ? Otherwise I'll just merge

tamuri:
Happy for you to merge.

PR #38, opened 2019-02-18
Using OrderedDict instead of dict to preserve document order of components and variables

MichaelClerx:
The Python 3.6 dict implementation already guarantees this, but the contract does _not_, as in, it may change in the future

https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -38,7 +38,7 @@ class Component(object):
         self.siblings: Set[str] = set()
         self.encapsulated: Set[str] = set()
 
-        self.variables: OrderedDict[str, Dict] = OrderedDict()
+        self.variables: Dict[str, Dict] = OrderedDict()
         self.equations: List[sympy.Eq] = []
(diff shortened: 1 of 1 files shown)


cellmlmanip/model.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, name: str, model: 'Model') -> None:
         self.siblings: Set[str] = set()
         self.encapsulated: Set[str] = set()
 
-        self.variables: Dict[str, Dict] = {}
+        self.variables: OrderedDict[str, Dict] = OrderedDict()
```
jonc125: I think you might want `Dict` instead of `OrderedDict` in the type, judging by https://docs.python.org/3/library/typing.html#typing.Dict. (In Python 3.7 you can use `typing.OrderedDict`...)

MichaelClerx: Cheers!

Issue #39, opened 2019-02-28
Add RDF processing


jonc125:
For looking up variables by annotation we need to process the RDF in a model as a graph, in the same way as the [cellml_metadata.py](https://chaste.cs.ox.ac.uk/trac/browser/chaste/trunk/python/pycml/cellml_metadata.py) file in pycml.

Other useful bits of pycml code are the methods `get_variable_by_oxmeta_name` and `get_variables_by_ontology_term` in [pycml.py](https://chaste.cs.ox.ac.uk/trac/browser/chaste/trunk/python/pycml/pycml.py#L440), which use the RDF methods to do variable lookups, and are themselves used by the model manipulation code as the high-level interface. (Though we should probably have `get_variable_by_ontology_term` instead of `get_variable_by_oxmeta_name`, and pass it a full URI not local name.)

- [ ] Probably we should add a `metadata.py` file to cellmlmanip containing essentially the same public interface as the pycml metadata file.
    - We don't need to be able to serialise RDF back out
    - It may be better to define a class and have cellmlmanip models contain an instance of this class which provides access just to that model's RDF, rather than having a separate class that has a map of models to graphs.
    - The call to `cellml_model.xml_xpath(u'*/cml:variable[@cmeta:id="%s"]' % var_id)` should instead call the existing get var by cmeta_id method.
    - Some methods may not be needed by Web Lab, only Chaste code gen; only copy stuff needed by `find_variables` in the first instance.
- [ ] We should also copy the .ttl version of the ontology over (require rdflib >= 4 in our dependencies)
    - Eventually we'll want to have a single source of truth for the Web Lab ontologies, but for now each project that needs it can have its own copy.
    - This is only needed for the `transitive_subjects` method which is part of the advanced interface used for arbitrary vector outputs
- [ ] We'll need some tests
    - Have a test model with some basic metadata, and check that each of our public methods give the expected results. We might be able to reuse an existing test model for this.
    - `test_simple_odes.cellml` has the following annotations (I've used the `oxmeta` prefix purely for brevity, and the `cmeta:id` on the right):
        - `oxmeta:time` on `time`
        - `oxmeta:sodium_reversal_potential` on `sv11`
        -  many others like this
        - `urn:test-ns#parameter_a1` on `a1`
    - So we could test (in order of implementing):
        - `get_variable_by_ontology_term(oxmeta + 'time')` gets you `time`
[...]

PR #40, opened 2019-02-28
Add metadata symbol lookups

MichaelClerx:
Part of #39

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -579,10 +579,10 @@ class Model(object):
         x = get_symbols_by_rdf(property, value)
         if len(x) == 0:
             raise ValueError(
-                'No match found for `' + str(property) '`: `' + value + '`.')
+                'No match found for `' + str(property) + '`: `' + value + '`.')
         elif len(x) > 1:
             raise ValueError(
-                'Multiple variables match `' + str(property) '`: `' + value +
+                'Multiple variables match `' + str(property) + '`: `' + value +
                 '`.')
 
     def get_symbols_by_rdf(self, property, value=None):
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -568,52 +568,61 @@ class Model(object):
 
         raise KeyError('No variable with cmeta id "%s" found.' % str(cmeta_id))
 
-    def get_symbol_by_rdf(self, property, value=None):
+    def get_symbol_by_ontology_term(self, namespace, local_name):
         """
-        Like :meth:`get_symbols_by_rdf` but only returns a single variable
-        symbol.
+        Searches the RDF graph for a variable annotated with the given
+        ``namespace:local_name`` and returns its symbol.
 
-        Raises a `KeyError` if no match is found, or a `ValueError` if multiple
-        matching variables are found.
+        Specifically, this method searches for a unique variable annotated with
+        predicate ``http://biomodels.net/biology-qualifiers/is`` and the object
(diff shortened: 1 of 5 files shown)


MichaelClerx:
@jonc125 @tamuri I think this is ready for review now?

cellmlmanip/model.py
```diff
@@ -567,6 +568,65 @@ def get_symbol_by_cmeta_id(self, cmeta_id):
 
         raise KeyError('No variable with cmeta id "%s" found.' % str(cmeta_id))
 
+    def get_symbol_by_ontology_term(self, namespace, local_name):
+        """
+        Searches the RDF graph for a variable annotated with the given
+        ``namespace:local_name`` and returns its symbol.
```
jonc125: `namespace` here isn't a prefix, so this doc is potentially confusing. I'd call the parameter `namespace_uri` and describe this as `{namespace_uri}local_name` which is a standard representation; similarly in the next para and error messages below.

cellmlmanip/model.py
```diff
@@ -567,6 +568,65 @@ def get_symbol_by_cmeta_id(self, cmeta_id):
...
+
+        Specifically, this method searches for a unique variable annotated with
+        predicate ``http://biomodels.net/biology-qualifiers/is`` and the object
+        specified by ``namespace + local_name``.
+
+        Will raise a ``KeyError`` if no variable with the given annotation is
+        found, and a ``RuntimeError`` if more than one variable with the given
```
jonc125: `RuntimeError` is a bit generic. Would `ValueError` be better?

cellmlmanip/rdf.py
```diff
@@ -0,0 +1,77 @@
...
+        # Create blank node
+        node = rdflib.BNode()
+
+    return node
+'''
+
+
+def create_rdf_node(namespace, local_name):
+    """
+    Creates and returns an RDF node from the given ``ns_``,
```
jonc125: Mismatched name

tests/test_rdf.py
```diff
@@ -0,0 +1,74 @@
...
+        model.get_symbol_by_ontology_term(oxmeta, 'time')
+    assert 'Multiple variables annotated with' in str(e)
+
+    # Annotation with a cmeta id that doesn't exist
+    with pytest.raises(KeyError) as e:
+        model.get_symbol_by_ontology_term(oxmeta, 'membrane_potential')
+    assert 'No variable with cmeta id' in str(e)
+
+    # Annotation of something that isn't a variable
+    # TODO: At the moment this fails because there is no variable called 'c'.
```
jonc125: I think failing like this is fine: it's correct to say that there isn't a variable with this id, after all!

cellmlmanip/model.py
MichaelClerx: I suppose... If it's legal to have multiple things annotated with "is", but not to request one and insist it's unique...

cellmlmanip/rdf.py
MichaelClerx: Thanks!

tests/test_rdf.py
MichaelClerx: ok!

cellmlmanip/model.py
jonc125: It's slightly tenuous, but I don't think would surprise users!

cellmlmanip/model.py
MichaelClerx: Done!

Issue #53, opened 2019-07-12
Installing dev dependencies fails


MichaelClerx:
```
$ virtualenv venv -p python
$ source venv/bin/activate
$ pip install -r requirements/dev.txt
```

results in

```
Building wheels for collected packages: lxml, mpmath
  Building wheel for lxml (setup.py) ... error
  ERROR: Complete output from command /home/michael/dev/weblab/cellmlmanip/venv/bin/python3 -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-msv25rs6/lxml/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-28phyf_8 --python-tag cp37:
  ERROR: Building lxml version 4.2.1.
  Building without Cython.
  ERROR: b'/bin/sh: xslt-config: command not found\n'
  ** make sure the development packages of libxml2 and libxslt are installed **
  
  Using build configuration of libxslt
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.7
  creating build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/doctestcompare.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/pyclasslookup.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/cssselect.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/usedoctest.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/__init__.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/ElementInclude.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/builder.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/_elementpath.py -> build/lib.linux-x86_64-3.7/lxml
  copying src/lxml/sax.py -> build/lib.linux-x86_64-3.7/lxml
  creating build/lib.linux-x86_64-3.7/lxml/includes
  copying src/lxml/includes/__init__.py -> build/lib.linux-x86_64-3.7/lxml/includes
  creating build/lib.linux-x86_64-3.7/lxml/html
[...]

PR #56, opened 2019-08-14
Fix issues with dev-requirements

MichaelClerx:
Fixes two issues:

- The pinned lxml version can no longer be installed via pip: #53
- The pip-tools version no longer works with newer versions of pip itself


MichaelClerx:
(Note the appveyor build is not configured properly yet, and so won't pass!)

Issue #55, opened 2019-08-14
Add appveyor testing on windows


MichaelClerx:
Immediate use case: @MauriceHendrix is developing on windows
Long-term use case: We'd like cellmlmanip/weblab_cg/etc. to be useful for as many users as possible

MichaelClerx:
I did a brief test on windows and did find some bugs, so implementing this will involve fixing those. @tamuri could you maybe help with this part before moving on from this project?

MichaelClerx:
Fixed in #59

PR #59, opened 2019-08-14
Set up appveyor testing

MichaelClerx:
See #55

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,7 @@
 branches:
   only:
     - master
+    - develop
 
 environment:
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,18 @@
+branches:
+  only:
+    - master
+
+environment:
+
+  matrix:
+    - PYTHON: "C:\\Python36-x64"
+
+install:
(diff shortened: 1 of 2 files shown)


MichaelClerx:
@tamuri no idea what's happening here! Very weird error message. @MauriceHendrix do you see the same if you test cellmlmanip on windows?

MichaelClerx:
```
        # Annotation with a cmeta id that doesn't exist
        with pytest.raises(KeyError) as e:
            model.get_symbol_by_ontology_term(oxmeta, 'membrane_potential')
>       assert 'No variable with cmeta id' in str(e)
E       AssertionError: assert 'No variable with cmeta id' in '<ExceptionInfo KeyError tblen=4>'
E        +  where '<ExceptionInfo KeyError tblen=4>' = str(<ExceptionInfo KeyError tblen=4>)
tests\test_rdf.py:58: AssertionError
```

MichaelClerx:
Think I found it

MauriceHendrix:
It could be my build but lxml==4.2.2 gives an error
The version currently used on weblab-cg is perfecly fine.
This is while using venv rather than conda.

PR #63, opened 2019-08-14
update packages to latest versions (for #31)

### LINKED ISSUE #31: Cellmlmanip is incompatible with newer versions of sympy

### LINKED ISSUE #61: lxml==4.2.2 gives an error wne building on windows under venv (rather than conda) can we increase version in dev.txt?
It could be my build but lxml==4.2.2 gives an error
The version currently used on weblab-cg is perfecly fine.
This is while using venv rather than conda.

Can we increase version in dev.txt? (currently weblab-cg states 4.3.4

_Originally posted by @MauriceHendrix in https://github.com/ModellingWebLab/cellmlmanip/pull/59#issuecomment-521312298_

tamuri:
Should fix #31 and fix #61.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ CellML loading and model equation manipulation
 Set up a virtual environment, and install requirements:
 ```sh
 conda create -n cellmlmanip python=3.6
-source activate cellmlmanip
+conda activate cellmlmanip
 pip install -r requirements/dev.txt
 pip install -e .
 ```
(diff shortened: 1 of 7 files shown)


tamuri:
Also updates lxml (see #61)

jonc125:
Tests pass and code looks OK to me. If it works for you @MauriceHendrix please merge :)

PR #70, opened 2019-08-28
Remove cmeta add preferred unit name

MauriceHendrix:
I looked into removing get_symbol_by_cmeta_id from model.py however it is used internally. I added a comment and renamed it to _get_symbol_by_cmeta_id. I also added a new method to units.py: add_preferred_custom_unit_name this emthod allows you to set a preferred name for equivalent units. For example if the model has mV defined but I was to get back "millivolt" instead

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -1,4 +1,4 @@
-"""Classes to represent a flattened CellML model and metadata about its variables"""
+﻿"""Classes to represent a flattened CellML model and metadata about its variables"""
 import logging
 from collections import OrderedDict
 from io import StringIO
@@ -461,7 +461,7 @@ class Model(object):
     def _get_symbol_by_cmeta_id(self, cmeta_id):
         """Searches the given graph and returns the symbol for the variable with the given cmeta_id.
         PLEASE NOTE this does NOT get the oxmeta tag to get that use
-        get_symbol_by_ontology_term("https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata",
+        get_symbol_by_ontology_term("https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#",
                                     "cytosolic_calcium_concentration")
         """
         for v in self.graph:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -458,12 +458,11 @@ class Model(object):
         # This should be unreachable
         raise ValueError('No free variable set in model.')  # pragma: no cover
 
-    def get_symbol_by_cmeta_id(self, cmeta_id):
+    def _get_symbol_by_cmeta_id(self, cmeta_id):
         """Searches the given graph and returns the symbol for the variable with the
-        given cmeta_id.
+        given cmeta_id. 
+        PLEASE NOTE this does NOT get the oxmeta tag to get that use get_symbol_by_ontology_term("https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata", "cytosolic_calcium_concentration")
         """
-        # TODO: Either add an argument to allow derivative symbols to be fetched, or
-        #      create a separate method for them.
         for v in self.graph:
             if self.graph.nodes[v].get('cmeta_id', '') == cmeta_id:
                 return v
@@ -521,7 +520,7 @@ class Model(object):
                 # TODO This should eventually be implemented
                 raise NotImplementedError(
                     'Non-local annotations are not supported.')
-            symbols.append(self.get_symbol_by_cmeta_id(uri[1:]))
+            symbols.append(self._get_symbol_by_cmeta_id(uri[1:]))
 
         return symbols
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -459,9 +459,10 @@ class Model(object):
         raise ValueError('No free variable set in model.')  # pragma: no cover
 
     def _get_symbol_by_cmeta_id(self, cmeta_id):
-        """Searches the given graph and returns the symbol for the variable with the
-        given cmeta_id. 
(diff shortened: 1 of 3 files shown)


cellmlmanip/model.py
```diff
@@ -458,12 +458,12 @@ def get_free_variable_symbol(self):
         # This should be unreachable
         raise ValueError('No free variable set in model.')  # pragma: no cover
 
-    def get_symbol_by_cmeta_id(self, cmeta_id):
-        """Searches the given graph and returns the symbol for the variable with the
-        given cmeta_id.
+    def _get_symbol_by_cmeta_id(self, cmeta_id):
+        """Searches the given graph and returns the symbol for the variable with the given cmeta_id.
+        PLEASE NOTE this does NOT get the oxmeta tag to get that use
+        get_symbol_by_ontology_term("https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata",
```
jonc125: Shouldn't the first argument here end in a `#`?

jonc125: Ah, I see `rdf.py` has a check for a missing `#` and adds it to avoid the most common user error! But our reference examples should definitely have it - the canonical URL for the ontology base ends in a `#`.

MauriceHendrix: I didn't touch this, just changed the function name

jonc125: You added the extra content in the docstring, and that's where the extra `#` is wanted.

tests/test_unit_conversion.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+import os
+import cellmlmanip
+
+OXMETA = "https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata"
```
jonc125: We should probably append a `#` here too so we're always being consistent about doing things properly ourselves!

tests/test_unit_conversion.py
```diff
@@ -0,0 +1,19 @@
...
+def model():
+    return cellmlmanip.load_model(os.path.join(os.path.dirname(__file__), 'cellml_files', "test_simple_odes.cellml"))
+
+
+def test_add_preferred_custom_unit_name(model):
+    time_var = model.get_symbol_by_ontology_term(OXMETA, "time")
+    assert str(model.units.summarise_units(time_var)) == "ms"
+    model.units.add_preferred_custom_unit_name('millisecond', [{'prefix': 'milli', 'units': 'second'}])
+    assert str(model.units.summarise_units(time_var)) == "millisecond"
+    model.units.add_preferred_custom_unit_name('millisecond', [{'prefix': 'milli', 'units': 'second'}])
```
jonc125: Why doing the same thing twice?

cellmlmanip/units.py
```diff
@@ -120,6 +120,22 @@ def add_custom_unit(self, units_name, unit_attributes):
 
             self._define_pint_unit(units_name, unit_definition)
 
+    def add_preferred_custom_unit_name(self, units_name, unit_attributes):
+        """Set a prefered name for all equivalent custom units. If it does not exist yet, also define a new Pint unit
```
jonc125: Can we follow https://www.python.org/dev/peps/pep-0257/ for docstrings please?

cellmlmanip/model.py
```diff
@@ -458,12 +458,12 @@ def get_free_variable_symbol(self):
         # This should be unreachable
         raise ValueError('No free variable set in model.')  # pragma: no cover
 
-    def get_symbol_by_cmeta_id(self, cmeta_id):
-        """Searches the given graph and returns the symbol for the variable with the
-        given cmeta_id.
+    def _get_symbol_by_cmeta_id(self, cmeta_id):
```
jonc125: I'm not sure about this change. On the one hand it's internal at present. On the other hand other users of cellmlmanip might well find 'get symbol by id' functionality useful (e.g. because they're not as ontology focused as we are) and it doesn't need to be 'hidden'. So my inclination is actually to keep it public. The extra documentation should be sufficient to avoid user confusion.

MauriceHendrix: ok

tests/test_unit_conversion.py
MauriceHendrix: because the code does something slightly different if the unit already exists. In short it is to guarantee 100% code coverage of the diff.

tests/test_unit_conversion.py
MauriceHendrix: done

tests/test_unit_conversion.py
jonc125: I'd add a comment to that effect then.

A stronger test case might be where the unit exists through another pathway than this new method?

MauriceHendrix: I don't think there is any other way to trigger the except within `add_preferred_custom_unit_name` other than adding the same unit twice

MauriceHendrix: comment added

cellmlmanip/units.py
MauriceHendrix: Can't see what's wrong with that

tests/test_unit_conversion.py
```diff
@@ -15,5 +15,8 @@ def test_add_preferred_custom_unit_name(model):
     assert str(model.units.summarise_units(time_var)) == "ms"
     model.units.add_preferred_custom_unit_name('millisecond', [{'prefix': 'milli', 'units': 'second'}])
     assert str(model.units.summarise_units(time_var)) == "millisecond"
+    ''' add_custom_unit does not allow adding already existing units but add_preferred_custom_unit_name does since we
```
jonc125: Using `# comment` style is probably more appropriate than docstring style here?

cellmlmanip/units.py
jonc125: The first line should be a brief one sentence summary of what the method does (probably just "Set a prefered name for all equivalent custom units." in this case). Then a blank line. Then paragraph-style further details.

tests/test_unit_conversion.py
MauriceHendrix: is it? from what i read unless they're on top they don't get treated as docstring

PR #73, opened 2019-09-03
for chaste code generation I would like to stay closer to order in th…

MauriceHendrix:
for chaste code generation I would like to stay closer to order in the model to stay closer to pycml output, therfore I needed a version of get_equations_for that does not do a lexicographical_sort

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -412,6 +412,7 @@ class Model(object):
 
     def get_equations_for(self, symbols, lexicographical_sort=True):
         """Get all equations for given collection of symbols
+        
         lexicographical_sort indicates whether the result will be sorted in lexicographical order"""
         graph = self.get_equation_graph()
         if lexicographical_sort:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -410,10 +410,14 @@ class Model(object):
         self.graph = graph
         return graph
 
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -410,10 +410,14 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
         self.graph = graph
         return graph
 
-    def get_equations_for(self, symbols):
-        """Get all equations for given collection of symbols"""
+    def get_equations_for(self, symbols, lexicographical_sort=True):
+        """Get all equations for given collection of symbols
```
MichaelClerx: Docstrings should have the format

    """
    One-line description

    Longer description, which may include a list of arguments (in a syntax that's unfortunately project-specific).
    The language for docstrings is "reStructuredText", in which you indicate variable names (and other code) with two backticks, like ``this``
    """

MichaelClerx:
What order would you like them in instead then? Make sure it's deterministic, so you always get the same output in tests

jonc125:
Pycml certainly ensures deterministic output. I can't recall what ordering it applies to equations beyond a topographical sort though.

MauriceHendrix:
> Needs better docstring and more unit tests --> right now cover is reduced

that's why this is a draft I'm working on that

PR #74, opened 2019-09-03
Unordered get equations for

MauriceHendrix:
Would like to be able to not get the get_equations_for results in lexicographical_topological order (but only in topological order

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -414,7 +414,8 @@ class Model(object):
         """Get all equations for given collection of symbols
 
         Results are sorted in topographical order.
-        Lexicographical_sort indicates whether the result is sorted in lexicographical order first
+        :param symbols: the symbols to get the equations for
+        :param lexicographical_sort: indicates whether the result is sorted in lexicographical order first
         """
         graph = self.get_equation_graph()
         if lexicographical_sort:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -412,8 +412,10 @@ class Model(object):
 
     def get_equations_for(self, symbols, lexicographical_sort=True):
         """Get all equations for given collection of symbols
-        
-        lexicographical_sort indicates whether the result will be sorted in lexicographical order"""
+
+        Results are sorted in topographical order.
+        Lexicographical_sort indicates whether the result is sorted in lexicographical order first
+        """
         graph = self.get_equation_graph()
         if lexicographical_sort:
             sorted_symbols = nx.lexicographical_topological_sort(graph, key=str)
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -412,6 +412,7 @@ class Model(object):
 
     def get_equations_for(self, symbols, lexicographical_sort=True):
         """Get all equations for given collection of symbols
+        
         lexicographical_sort indicates whether the result will be sorted in lexicographical order"""
         graph = self.get_equation_graph()
         if lexicographical_sort:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -410,10 +410,14 @@ class Model(object):
         self.graph = graph
         return graph
 
-    def get_equations_for(self, symbols):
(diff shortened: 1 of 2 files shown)


MauriceHendrix:
> Looks fine to me. But are we sure this will match the PyCml ordering?

For what I'm working on and the models I've tried it seems like the reverse of topological order matches what pycml uses but I don't understand pycml code nearly well enough to know if that's genericly the case. 

(Also I didn't add the reverse to cellmlmanip as I thought it would make it confusing if the flag also had the side effect of reversing the output)

cellmlmanip/model.py
```diff
@@ -410,10 +410,17 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
         self.graph = graph
         return graph
 
-    def get_equations_for(self, symbols):
-        """Get all equations for given collection of symbols"""
+    def get_equations_for(self, symbols, lexicographical_sort=True):
+        """Get all equations for given collection of symbols
+
+        Results are sorted in topographical order.
+        Lexicographical_sort indicates whether the result is sorted in lexicographical order first
```
jonc125: I think we're using `:param lexicographical_sort: indicates...` elsewhere?

tests/test_hodgkin.py
```diff
@@ -240,3 +242,21 @@ def test_evaluation(self, graph, model):
             expected = evaluated_derivatives[state_symbol.name]
             actual = evaluated_deriv
             assert float(actual) == pytest.approx(expected)
+
+    def test_get_symbol_by_ontology_term(self, graph, model):
```
jonc125: Why is this unrelated test in the PR? (Not that I have an issue with the test itself!)

tests/test_hodgkin.py
```diff
@@ -240,3 +242,21 @@ def test_evaluation(self, graph, model):
...
+        membrane_voltage_var = model.get_symbol_by_ontology_term(OXMETA, "membrane_voltage")
+        assert (isinstance(membrane_voltage_var, sympy.symbol.Dummy))
+        assert str(membrane_voltage_var) == "_membrane$V"
+
+    def test_get_equations_for(self, graph, model):
+        # Get equations for membrane_fast_sodium_current both ordered and unordered
+        membrane_fast_sodium_current = model.get_symbol_by_ontology_term(OXMETA, "membrane_fast_sodium_current")
+        equations = model.get_equations_for([membrane_fast_sodium_current])
+        unordered_equations = model.get_equations_for([membrane_fast_sodium_current], False)
+        # There should be 4 in this model
```
jonc125: Should we also check that the equations are as expected? Or is that tested elsewhere?

tests/test_hodgkin.py
```diff
@@ -240,3 +242,21 @@ def test_evaluation(self, graph, model):
             expected = evaluated_derivatives[state_symbol.name]
             actual = evaluated_deriv
             assert float(actual) == pytest.approx(expected)
+
+    def test_get_symbol_by_ontology_term(self, graph, model):
+        membrane_voltage_var = model.get_symbol_by_ontology_term(OXMETA, "membrane_voltage")
+        assert (isinstance(membrane_voltage_var, sympy.symbol.Dummy))
```
jonc125: You don't need () around assertions; similarly below.

tests/test_hodgkin.py
MauriceHendrix: because it's a (previously untested) method that the other test uses, the other test is needed to ensure 100% coverage and I thought it would be neater to also have a test for this method rather than relying on it failing in the test for get_equations_for

cellmlmanip/model.py
MauriceHendrix: good point although shouldn't symbols also have that?

tests/test_hodgkin.py
MauriceHendrix: ok, fixed

tests/test_hodgkin.py
MauriceHendrix: Good point to be fair it was just laziness :) added now.

PR #78, opened 2019-09-12
Investigate units

skeating:
Working on units

* [x] Make sure the unit calculator deals with all math
* [x] throw exceptions rather than return None when units cannot be calculated

### DIFF
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -138,7 +138,7 @@ class BooleanUnitsError(UnitError):
 
     def __init__(self, expression):
         self.expression = expression
-        self.message = 'This expression involves boolean values which do not conform to' \
+        self.message = 'This expression involves boolean values which do not conform to ' \
                        'unit dimensionality rules.'
 
 
@@ -488,6 +488,12 @@ class UnitCalculator(object):
             logger.critical('Boolean return: %s', expr)
             raise BooleanUnitsError('%s' % expr)
 
+        elif expr.is_Boolean:
+            # following discussion with Michael we decided that since a
+            # variable in cellml can never have a boolean value then
+            # we should not encounter expression that return booleans
+            raise BooleanUnitsError('%s' % expr)
+
         elif expr.is_Function:
             # List of functions that have been checked
             if str(expr.func) not in ['cos', 'acos', 'exp', 'floor', 'log', 'Abs', 'tanh', 'ceiling',
@@ -540,6 +546,12 @@ class UnitCalculator(object):
                 return 1 * self.ureg.dimensionless
         elif expr == sympy.pi:
             return math.pi * self.ureg.dimensionless
+        elif expr == sympy.E:
+            return math.e * self.ureg.dimensionless
+        elif expr == sympy.oo:
+            return math.inf * self.ureg.dimensionless
+        elif expr == sympy.nan:
+            return math.nan * self.ureg.dimensionless
         elif expr.is_Derivative:
             out = quantity_per_arg[0] / quantity_per_arg[1]
             return out
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -99,8 +99,10 @@ class UnitStore(object):
 
     def add_custom_unit(self, units_name, unit_attributes):
         """Define a new Pint unit definition to the unit registry
-        :param units_name:
-        :param unit_attributes:
+        :param units_name: name of unit to add - cannot be an existing CellML unit
+        :param unit_attributes: dict object for unit
+            {'multiplier': float, 'units': string, 'exponent': integer, 'prefix': string/integer}
+            Not all fields necessary but 'units' must match a unit in Pint registry
         """
         assert units_name not in CELLML_UNITS, 'Cannot redefine CellML unit <%s>' % units_name
 
@@ -122,13 +124,17 @@ class UnitStore(object):
 
     def add_base_unit(self, units_name):
         """Define a new base unit in the Pint registry
-        :param units_name:
+        :param units_name: string name of unit to add
         """
         assert units_name not in CELLML_UNITS, 'Cannot redefine CellML unit <%s>' % units_name
         assert not self._is_unit_defined(units_name), 'Unit <%s> already exists' % units_name
         self._define_pint_unit(units_name, '{name}=[{name}]'.format_map({'name': units_name}))
 
     def _is_unit_defined(self, name):
+        """ Check whether the unit already exists in Pint registry
+        :param name: string name of the unit
+        :return: True if exists, else False
+        """
         try:
             getattr(self.ureg, name)
             return True
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -79,6 +79,58 @@ TRIG_FUNCTIONS = {
 }
 
 
+class UnitError(Exception):
+    """ base class for unit errors
+    """
+    pass
+
+
+class InvalidUnitsError(UnitError):
+    """ generic invalid units error
+    @param expression -- input expression in which the error occurred
+    """
+
+    def __init__(self, expression):
+        self.expression = expression
+        self.message = 'The units of this expression cannot be calculated.'
+
+
+class InputArgumentsInvalidUnitsError(UnitError):
+    """ error when the arguments to a function have incorrect units
+    @param expression -- input expression in which the error occurred
+    """
+
+    def __init__(self, expression):
+        self.expression = expression
+        self.message = 'The arguments to this expression should all have the same units.'
+
+
+class InputArgumentsMustBeDimensionlessError(UnitError):
+    """ error when the arguments to a function have incorrect units when they should be dimensionless
+    @param expression -- input expression in which the error occurred
+    """
+
+    def __init__(self, expression, position=''):
+        self.expression = expression
+        if position.__len__() == 0:
+            self.message = 'The arguments to this expression should be dimensionless.'
+        else:
+            self.message = 'The %s argument to this expression should be dimensionless.' % position
+
+
+class InputArgumentMustBeNumberError(UnitError):
+    """ error when the input argument must be a number
+    @param expression -- input expression in which the error occurred
+    """
+
+    def __init__(self, expression, position=''):
+        self.expression = expression
+        if position.__len__() == 0:
+            self.message = 'The argument to this expression should be a number.'
+        else:
+            self.message = 'The %s argument to this expression should be a number.' % position
+
+
 class UnitStore(object):
     """Wraps the underlying Pint UnitRegistry to provide unit handling for the model's Sympy
     expressions. Getting and checking units is handled by this class.
@@ -307,7 +359,7 @@ class UnitCalculator(object):
         # if there was an error determining the quantity of an argument
         if None in quantity_per_arg:
             # error
-            return None
+            raise InvalidUnitsError('%s' % expr)
 
         # Terminal atoms in expressions (Integers and Rationals are used by Sympy itself)
         if expr.is_Symbol:
@@ -361,13 +413,13 @@ class UnitCalculator(object):
             # exponent must be dimensionless
             if exponent.units != self.ureg.dimensionless:
                 logger.critical('Exponent of pow is not dimensionless %s', expr)
-                return None
+                raise InputArgumentsMustBeDimensionlessError('%s' % expr, 'second')
 
             if not isinstance(exponent.magnitude, (sympy.Number, numbers.Number)):
                 logger.critical('Exponent of pow is not a number (is %s): %s',
                                 type(exponent.magnitude).__name__,
                                 expr)
-                return None
+                raise InputArgumentMustBeNumberError('%s' % expr, 'second')
 
             # if base is dimensionless, return is dimensionless
             if base.units == self.ureg.dimensionless:
@@ -389,7 +441,7 @@ class UnitCalculator(object):
                 return out
 
             logger.warning('Add args do not have the same unit: %s', expr)
-            return None
+            raise InputArgumentsInvalidUnitsError('%s' % expr)
 
         elif expr.is_Piecewise:
             # If unit of each expression in piecewise is the same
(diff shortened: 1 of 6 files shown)


skeating:
A couple of questions relating to UnitCalculator that might need @jonc125 and/or @tamuri to answer:

1. units.py 404
        elif expr.is_Function:
            # List of functions that have been checked
            # SK: I'm not sure that this is accurate - we handle any trig function
            # but until I added one to tests none were tested so - not sure what we mean by checked
            if str(expr.func) not in ['cos', 'acos', 'exp', 'floor', 'log', 'Abs', 'tanh']:
                logger.warning('Have not checked unit arithmetic for function %s', expr.func)

Presumably a line from when creating code or does 'checked' have a deeper meaning

2. units 437
            elif expr.func == sympy.exp:
                # requires operands to have units of dimensionless.
                # result of these has units of dimensionless.
                if self._is_dimensionless(quantity_per_arg[0]):
                    # is the operand is a float
                    if isinstance(quantity_per_arg[0].magnitude, float):
                        # return the exponential of the float as dimensionless
                        return self.ureg.Quantity(math.exp(quantity_per_arg[0].magnitude),
                                                  self.ureg.dimensionless)

                    # magnitude contains an unresolved symbol, we lose it here!
                    # we don't lose it - the unit will be dimensionless - we are just not able to
                    # determine the magnitude of the result    <====== is this what you meant by 'lose it'
                    return 1 * self.ureg.dimensionless

3. units 420
            elif expr.func == sympy.log:
                if self._is_dimensionless(quantity_per_arg[0]):
                    return 1 * self.ureg.dimensionless
                # SK: why raise an exception here when anything else that fails returnS None
                raise ValueError('log args not dimensionless (%s)' %
                                 [x.units for x in quantity_per_arg])

Is this a left over from implementation - or do we want to throw exceptions rather than return None

tamuri:
> # but until I added one to tests none were tested so - not sure what we mean by checked

tamuri:
1. > but until I added one to tests none were tested so - not sure what we mean by checked

The condition handles any Sympy Function but I [manually] checked the unit arithmetic for those listed functions. There might be other functions that will need custom code (as has been written for Abs, floor, exp etc.)

2. > determine the magnitude of the result    <====== is this what you meant by 'lose it'

Yes, exactly, the magnitude of the quantity is lost.

3. > Is this a left over from implementation - or do we want to throw exceptions rather than return None

You're right- error handling needs to be improved & consistent.

jonc125:
Which way we become consistent depends on what turns out to be nicest for the consumer.

skeating:
Main changes in this PR is that 
1. UnitCalculator::traverse function now handles all math that is might expect from a cellml model
2. it throws exceptions if it cannot work out units - with a range of possible exceptions to hopefully be as clear as possible.

My slight issue is that the traverse functions returns a Quantity that has a unit and a magnitude. In some cases the magnitude will be correct - in others it will not. Should I
a) try and make it as accurate as possible
b) put Nan as magnitude for all
c) do a) and then b) for impossible things

Don't know how urgent this would be.

jonc125:
I've put some notes on Slack for where this ultimately gets used in applying protocols, code gen etc. Hopefully it helps as a starting point! There are possibly only a couple of cases where we need to determine units for a full expression, with scaling factor, if we don't worry about validating model units in general.

cellmlmanip/units.py
```diff
@@ -385,11 +499,29 @@ def traverse(self, expr: sympy.Expr):
...
+        elif expr.is_Boolean:
+            # following discussion with Michael we decided that since a
+            # variable in cellml can never have a boolean value then
+            # we should not encounter expression that return booleans
+            raise BooleanUnitsError('%s' % expr)
 
         elif expr.is_Function:
             # List of functions that have been checked
-            if str(expr.func) not in ['cos', 'acos', 'exp', 'floor', 'log', 'Abs', 'tanh']:
+            if str(expr.func) not in ['cos', 'acos', 'exp', 'floor', 'log', 'Abs', 'tanh', 'ceiling',
```
MauriceHendrix: Shouldn't that be Pow? I tried pow with lowercase the other day in the interpreter and it didn't work but Pow did

skeating:
Adding tests with a model with bad units shows that this can be covered up by the parser; so should my next priority be #80 or #84

skeating:
Cannot get diff coverage higher 👎

jonc125:
> Cannot get diff coverage higher 👎

It's still improving the overall coverage though! So I'm not too concerned at that.

jonc125:
> Adding tests with a model with bad units shows that this can be covered up by the parser; so should my next priority be #80 or #84

My inclination would be to focus on issues/features that are blockers for the next steps on Chaste code gen and/or WL protocol parsing, rather than more low-level tests.

tests/test_model_bad_units.py
```diff
@@ -0,0 +1,44 @@
...
+        model.get_equation_graph(True)  # set up the graph - it is not automatic
+        symbol_a = model.get_symbol_by_cmeta_id("a")
+        symbol_b = model.get_symbol_by_cmeta_id("b")
+        equation = model.get_equations_for([symbol_b])
+        assert len(equation) == 2
+        assert equation[0].lhs == symbol_a
+        assert equation[0].rhs == 2.0
+        assert model.units.summarise_units(equation[0].lhs) == 'ms'
+        # cellml file states a (ms) = 1 (ms) + 1 (second)
+        # this does not raise and error
```
jonc125: Yeah, that's actually consistent with the spec bizarrely (although pycml will flag it up) - processing s/w is only required to check connections. And iirc (@MichaelClerx will know for sure) in CellML 2 s/w *shouldn't* check inside equations!

cellmlmanip/units.py
```diff
@@ -303,26 +303,24 @@ def summarise_units(self, expr: sympy.Expr):
...
-            logger.fatal('-> %s', printer.doprint(expr))
-            return None
-        except KeyError as error:
-            printer = ExpressionWithUnitPrinter(symbol_info=self.model.dummy_metadata)
-            logger.fatal('Could not summaries units: %s', expr)
-            logger.fatal(error.message)
-            logger.fatal('-> %s', printer.doprint(expr))
-            return None
+        found = unit_calculator.traverse(expr)
+        # except UnitError as error:
```
jonc125: Should the commented stuff be removed?

cellmlmanip/units.py
```diff
@@ -362,30 +372,43 @@ def traverse(self, expr: sympy.Expr):
...
+        # need to work out units of each child atom of the expression
+        # piecewise and derivatives are special cases
         if expr.is_Piecewise:
-            # Treat each sub-function an argument (& ignore the condition)
+            # Treat each sub-function an argument (& ignore the conditional statements)
             for arg in expr.args:
                 quantity_per_arg.append(self.traverse(arg[0]))
         elif expr.is_Derivative:
+            # In a derivative expression the first argument is function to be differentiated
+            # the second argument is a tuple of which the first is symbol of the differentiator(?)
```
jonc125: How bizarre!

tamuri: Yup! 🤷‍♂ Introduced in latest version of Sympy; second argument of second tuple is the _n_ for nth derivatives.

cellmlmanip/units.py
```diff
@@ -79,6 +79,76 @@
 }
 
 
+class UnitError(Exception):
+    """ base class for unit errors
+    """
+    pass
+
+
+class InvalidUnitsError(UnitError):
```
MichaelClerx: Better description? "Invalid" in some sense that's not already implied by "Error"

cellmlmanip/units.py
```diff
@@ -281,29 +367,45 @@ def traverse(self, expr: sympy.Expr):
...
+                 InputArgumentsMustBeDimensionlessError - if input arguments should be dimensionless
+                 InputArgumentsInvalidUnitsError - if input arguments should have same units
+                 InputArgumentMustBeNumberError - if one of input arguments should be a number
+        NOTE: Sympy will throw exceptions if the expression is badly formed before we even get here
         """
         # collect the units for each argument of the expression (might be sub-expression)
         quantity_per_arg = []
 
-        # If the expression is a piecewise
+        if expr.is_Matrix:
```
MichaelClerx: Is this not caught by the generic case at the bottom?

(Would up your test coverage if it was and you could remove this, by the way, at the moment the final `else` is uncovered)

cellmlmanip/units.py
```diff
@@ -417,24 +553,40 @@ def traverse(self, expr: sympy.Expr):
...
         elif expr.is_Derivative:
             out = quantity_per_arg[0] / quantity_per_arg[1]
             return out
 
+        # constants in cellml that are all specified as dimensionless
+        elif expr == sympy.pi:
+            return math.pi * self.ureg.dimensionless
+        elif expr == sympy.E:
+            return math.e * self.ureg.dimensionless
+        elif expr == sympy.oo:
```
MichaelClerx: oo and nan need unit tests

cellmlmanip/units.py
skeating: Weird one. No, if you create a matrix expression - so expr.is_matrix is True then expr.is_Piecewise does not exist and code crashes.

I put this in but did then stop testing the other edge cases as we shouldnt encounter them from cellml files anyway.

cellmlmanip/units.py
skeating: I have

skeating: # constants
        assert unit_calculator.traverse(sp.pi).units == ureg.dimensionless
        assert unit_calculator.traverse(sp.E).units == ureg.dimensionless
        assert unit_calculator.traverse(sp.oo).units == ureg.dimensionless
        assert unit_calculator.traverse(sp.nan).units == ureg.dimensionless

skeating: but coverage claims they are not tested !!!

tests/test_model_bad_units.py
skeating: I added an new issue #90 where we can discuss what we did with this.

cellmlmanip/units.py
```diff
@@ -322,13 +424,22 @@ def traverse(self, expr: sympy.Expr):
...
-            out = int(expr) * self.ureg.dimensionless
-            return out
-        elif expr.is_Rational:
-            # NOTE: can't send back Rational(1,2) * u.dimensionless
-            # Used by Sympy for root e.g. sqrt(x) == Pow(x, Rational(1,2))
-            out = float(expr) * self.ureg.dimensionless
+
+        elif expr.is_Number:
+            units = self.ureg.dimensionless
+            for key in self.dummy_metadata:
```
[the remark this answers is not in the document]
skeating: the dummy_metadata is the symbol dictionary created by reading in a model. So it has entries
{key_symbol: MetaDummyObject} for each symbol but it is the number field of the MetaDummy object that actually matches the expression. The key is whatever symbol the reader decided to use. ie. it might replace 'a' in cellml with 'symbol_a'.

So I need to loop thru each entry in the symbol dict and then see if it is the matching number in order to get the units.

e.g. the tests set up
        symbol_info = { (snipped)
            a: MetaDummy('a', ureg.meter, a),
           _1: MetaDummy('_1', ureg.kelvin, _1, number=sp.Float(1.0)),
        }

If you can suggest a better loop please do.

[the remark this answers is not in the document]
MichaelClerx: OK. Can we change it to `for key, metadata in self.dummy_metadata.items()` at least? I'll make an issue for the weird keys

cellmlmanip/units.py
MauriceHendrix: Codecov is correct.
I looked at it and it seems that both
`sp.oo.is_Number == True`
`sp.nan.is_Number == True`

Counter intuitive I know, but it can be fixed easily by moving those 2 checks before the is-number check I think.

cellmlmanip/units.py
```diff
@@ -398,14 +527,21 @@ def traverse(self, expr: sympy.Expr):
             elif expr.func == sympy.floor:
                 # if we're flooring a sympy expression
                 if isinstance(quantity_per_arg[0].magnitude, sympy.Expr):
-                    # return 1...?
                     return 1 * quantity_per_arg[0].units
                 return math.floor(quantity_per_arg[0].magnitude) * quantity_per_arg[0].units
+            elif expr.func == sympy.ceiling:
+                # if we're ceiling a sympy expression
+                if isinstance(quantity_per_arg[0].magnitude, sympy.Expr):
+                    return 1 * quantity_per_arg[0].units
```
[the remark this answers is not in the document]
MichaelClerx: OK I know this know, it's meant to return a quantity not a unit :D

cellmlmanip/units.py
```diff
@@ -385,11 +496,29 @@ def traverse(self, expr: sympy.Expr):
                 return out
```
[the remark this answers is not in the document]
skeating: I'm calling this resolved for this issue and transferred to #39 for more thoughts.

cellmlmanip/units.py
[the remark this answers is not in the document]
skeating: means it returns a quantity not just unit

cellmlmanip/units.py
```diff
@@ -79,6 +79,77 @@
 }
 
 
+class UnitError(Exception):
+    """ base class for unit errors
```
MichaelClerx: I'd prefer it if we added capitalisation and punctuation in docstrings, so that we can easily generate nice API documentation later (e.g. via readthedocs.io, like here: https://pints.readthedocs.io/en/latest/ )

jonc125:
Looks like there are some legacy issues with docstrings generally in cellmlmanip :)

cellmlmanip/units.py
skeating: Done for units.py which is the main file I touched here. Will do a check as I hit other files.

MichaelClerx: Thanks so much! I'll make an issue for the other ones

PR #85, opened 2019-09-16
added input order preservation and exclusion list to get_equations_for

MauriceHendrix:
I have had a play with get_equations_for and it's not straightforward to preserve an order resembling the input list due to how the digraph works. 
This way does seem seem to work, although it feels like it can be written up nicer maybe. Specifically it keeps also the original sorting as the default option (hence only acts differently if you provide a flag), not sure if that's useful.

Unfortunately `graph.predecessors(parent)` doesn't always return things in the same order, hence the sorting. 

Also: I'm assuming that if I get predecessors all those don't have any interdependence. I.e. one of the predecessors shouldn't contain another one, as then it should be a parent? If that's not necessarily the case, we would need a more complex sorting.

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -7,6 +7,7 @@ from typing import Dict, List, Union
 import networkx as nx
 import rdflib
 import sympy
+from collections import deque
 
 from cellmlmanip.rdf import create_rdf_node
 from cellmlmanip.units import UnitStore
@@ -410,27 +411,49 @@ class Model(object):
         self.graph = graph
         return graph
 
-    def get_equations_for(self, symbols, lexicographical_sort=True):
+    def get_equations_for(self, symbols, sort_by_input_symbols=False, excluded_symbols=[]):
         """Get all equations for given collection of symbols
 
         Results are sorted in topographical order.
         :param symbols: the symbols to get the equations for
-        :param lexicographical_sort: indicates whether the result is sorted in lexicographical order first
+        :param sort_by_input_symbols: indicates whether the sorting in symbols determines the sorting of the output
+        :param excluded_symbols: list of symbols to exlcude
         """
         graph = self.get_equation_graph()
-        if lexicographical_sort:
-            sorted_symbols = nx.lexicographical_topological_sort(graph, key=str)
-        else:
-            sorted_symbols = nx.topological_sort(graph)
 
-        # Create set of symbols for which we require equations
-        required_symbols = set()
+        def get_parents(node, graph=graph):
+            parents = []
+            to_process = deque()
+            to_process.appendleft(node)
+            while len(to_process) > 0:
+                parent = to_process.popleft()
+                ancestors = [a for a in graph.predecessors(parent)]
+                ancestors.sort(key=str)
+                for a in ancestors:
+                    if a not in parents:
+                        to_process.appendleft(a)
+                if parent not in parents:
+                    parents.insert(0, parent)
+            return parents
+
+        # Create list of symbols for which we require equations (list instead of set, in order to preserve order)
+        required_symbols = []
         for output in symbols:
-            required_symbols.add(output)
-            required_symbols.update(nx.ancestors(graph, output))
+            if output not in required_symbols and output not in excluded_symbols:
+                for p in get_parents(output):
+                    if p not in required_symbols:
+                        required_symbols.append(p)
+                if output not in required_symbols:
+                    required_symbols.append(output)
 
         eqs = []
-        for symbol in sorted_symbols:
+
+        if sort_by_input_symbols:
+            to_search_in = required_symbols
+        else:
+            to_search_in = nx.lexicographical_topological_sort(graph, key=str)
+
+        for symbol in to_search_in:
             # Ignore symbols we don't need
             if symbol not in required_symbols:
                 continue
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -439,7 +439,8 @@ class Model(object):
         for output in symbols:
             if output not in required_symbols and output not in excluded_symbols:
                 for p in get_parents(output):
-                    required_symbols.append(p)
+                    if p not in excluded_symbols and p not in required_symbols:
+                        required_symbols.append(p)
 
         eqs = []
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -430,8 +430,7 @@ class Model(object):
                 ancestors.sort(key=str)
                 for a in ancestors:
                     to_process.appendleft(a)
-                if parent not in parents:
-                    parents.insert(0, parent)
+                parents.insert(0, parent)
             return parents
 
         # Create list of symbols for which we require equations (list instead of set, in order to preserve order)
@@ -439,8 +438,7 @@ class Model(object):
         for output in symbols:
             if output not in required_symbols and output not in excluded_symbols:
                 for p in get_parents(output):
-                    if p not in required_symbols:
-                        required_symbols.append(p)
+                    required_symbols.append(p)
 
         eqs = []
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -441,8 +441,6 @@ class Model(object):
                 for p in get_parents(output):
                     if p not in required_symbols:
                         required_symbols.append(p)
-                if output not in required_symbols:
-                    required_symbols.append(output)
 
         eqs = []
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -419,17 +419,20 @@ class Model(object):
         :param excluded_symbols: list of symbols to exlcude
         """
         graph = self.get_equation_graph()
+        sorted_symbols = [s for s in nx.topological_sort(graph)]
 
-        def get_parents(node, graph=graph):
+        def get_parents(node, graph=graph, sorted_symbols=sorted_symbols):
             parents = []
             to_process = deque()
             to_process.appendleft(node)
             while len(to_process) > 0:
                 parent = to_process.popleft()
                 ancestors = [a for a in graph.predecessors(parent)]
-                ancestors.sort(key=str)
-                for a in ancestors:
-                    to_process.appendleft(a)
+                for sym in sorted_symbols:
+                    if sym in ancestors:
+                        #ancestors.sort(key=str)
+                        #for a in ancestors:
+                        to_process.appendleft(sym)
                 parents.insert(0, parent)
             return parents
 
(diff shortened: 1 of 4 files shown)


cellmlmanip/model.py
```diff
@@ -410,27 +410,44 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
         self.graph = graph
         return graph
 
-    def get_equations_for(self, symbols, lexicographical_sort=True):
+    def get_equations_for(self, symbols, sort_by_input_symbols=False, excluded_symbols=[]):
```
skeating: Normally I would say it was not ideal to change the function signature - but since it is entirely internal I think its okay.

tests/test_hodgkin.py
```diff
@@ -276,34 +276,40 @@ def test_has_ontology_annotation(self, model):
...
+        membrane_stimulus_current_offset = model.get_symbol_by_ontology_term(OXMETA, "membrane_stimulus_current_offset")
+        membrane_stimulus_current_period = model.get_symbol_by_ontology_term(OXMETA, "membrane_stimulus_current_period")
+
+        equations = model.get_equations_for([membrane_fast_sodium_current, membrane_stimulus_current_offset])
+        input_ordered_equations = model.get_equations_for(
+            [membrane_fast_sodium_current, membrane_stimulus_current_offset,
+             membrane_stimulus_current_offset, membrane_stimulus_current_period],
+            True, [membrane_stimulus_current_period])
+
+        # There should be 10 in this model
```
skeating: 10 or 5 ?

tests/test_hodgkin.py
```diff
@@ -276,34 +276,40 @@ def test_has_ontology_annotation(self, model):
...
         ref_eq = [sympy.Eq(sympy.Dummy('membrane$E_R'), sympy.numbers.Float(-75.0)),
+                  sympy.Eq(sympy.Dummy('membrane$stim_start'), sympy.numbers.Float(10.0)),
                   sympy.Eq(sympy.Dummy('sodium_channel$E_Na'),
                            sympy.add.Add(sympy.Dummy('membrane$E_R'), sympy.numbers.Float(115.0))),
                   sympy.Eq(sympy.Dummy('sodium_channel$g_Na'), sympy.numbers.Float(120.0)),
                   sympy.Eq(sympy.Dummy('sodium_channel$i_Na'),
                            sympy.Dummy('sodium_channel_m_gate$m') ** 3.0 * sympy.Dummy('sodium_channel$g_Na') *
                            sympy.Dummy('sodium_channel_h_gate$h') * (sympy.Dummy('membrane$V') -
-                           sympy.Dummy('sodium_channel$E_Na')))
-                  ]
```
skeating: I would leave the close bracket on a separate line as it makes it much clearer when doing a diff what has actually changed

tests/test_hodgkin.py
```diff
@@ -276,34 +276,40 @@ def test_has_ontology_annotation(self, model):
...
-        equations = model.get_equations_for([membrane_fast_sodium_current])
-        unordered_equations = model.get_equations_for([membrane_fast_sodium_current], False)
-        # There should be 4 in this model
-        assert len(equations) == len(unordered_equations) == 4
+        membrane_stimulus_current_offset = model.get_symbol_by_ontology_term(OXMETA, "membrane_stimulus_current_offset")
+        membrane_stimulus_current_period = model.get_symbol_by_ontology_term(OXMETA, "membrane_stimulus_current_period")
+
+        equations = model.get_equations_for([membrane_fast_sodium_current, membrane_stimulus_current_offset])
+        input_ordered_equations = model.get_equations_for(
+            [membrane_fast_sodium_current, membrane_stimulus_current_offset,
```
skeating: should membrane_stimulus_current_offset be here twice

cellmlmanip/model.py
```diff
@@ -410,27 +410,44 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
...
         """
         graph = self.get_equation_graph()
-        if lexicographical_sort:
-            sorted_symbols = nx.lexicographical_topological_sort(graph, key=str)
-        else:
-            sorted_symbols = nx.topological_sort(graph)
 
-        # Create set of symbols for which we require equations
-        required_symbols = set()
+        def get_parents(node, graph=graph):
```
skeating: would this function not be useful as a global function ?

tests/test_hodgkin.py
MauriceHendrix: it's there twice to check that this doesn't lead to duplication of equations even if the input list perhaps contains duplicates.
not 100% sure it's necessary to be fair.

tests/test_hodgkin.py
MauriceHendrix: ok

tests/test_hodgkin.py
MauriceHendrix: 5 sorry

cellmlmanip/model.py
MauriceHendrix: Perhaps I didn't  really look at the whole library. If you think it's useful feel free to put it elsewhere. This pull request was just me experimenting.

The main thing I'm wondering is whether it's correct. Are there ever any case where direct ancestors have an interdependence (e.g. x=5y, y=5)? (If so order is important if I'm able to generate C++ code)

cellmlmanip/model.py
MauriceHendrix: Yes good idea feel free to change :)

cellmlmanip/model.py
jonc125: Not sure what you mean by direct ancestors having an interdependence?

MauriceHendrix: ok so lets say I have `x = 5*y + 6*z`
ancestors:
y and z
 would it be possible for y and z then t have their own equations for example:
`y =2*z
z = 1`

jonc125: Almost certainly yes!

MauriceHendrix: So we somehow need to make sure z ends up before y in the example somehow

MauriceHendrix: Would a topological sort work?

jonc125: I thought get_equations_for was doing a topological sort? Using the nx graph library? Certainly that's the correct approach.

The trickier bit is determining a consistent relative ordering where there isn't a direct dependency.

MauriceHendrix: It does but not with my flag. i've changed that now

tests/test_hodgkin.py
skeating: If there is a possibility that duplication in the input list would cause duplication of equations; then this is really a separate test. I would do the call with no duplication as well and also check list with duplication produces the correct result

MauriceHendrix: ok added.

MauriceHendrix:
I just added code to prevent unwanted duplicates. Had to also add test so put it in test_simple_model_units.py (may need to change) and pull in simple_model_units.cellml from the  investigate units branch.
Also split out test for duplicate input in test_hodgkin.py

cellmlmanip/model.py
```diff
@@ -410,27 +410,46 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
...
-            required_symbols.update(nx.ancestors(graph, output))
+            if output not in required_symbols and output not in excluded_symbols:
+                for p in get_parents(output):
+                    if p not in excluded_symbols and p not in required_symbols:
+                        required_symbols.append(p)
 
         eqs = []
-        for symbol in sorted_symbols:
+
+        if sort_by_input_symbols:
```
jonc125: If this is false, we're doing a lot of work above that's not needed!

cellmlmanip/model.py
```diff
@@ -410,27 +410,46 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
...
-        if lexicographical_sort:
-            sorted_symbols = nx.lexicographical_topological_sort(graph, key=str)
-        else:
-            sorted_symbols = nx.topological_sort(graph)
-
-        # Create set of symbols for which we require equations
-        required_symbols = set()
+        sorted_symbols = [s for s in nx.topological_sort(graph)]
+
+        def get_parents(node, graph=graph, sorted_symbols=sorted_symbols):
```
jonc125: I'd like this at least to have a docstring, even if we don't make it a class method.

cellmlmanip/model.py
```diff
@@ -410,27 +410,46 @@ def get_equation_graph(self, refresh=False) -> nx.DiGraph:
...
         return graph
 
-    def get_equations_for(self, symbols, lexicographical_sort=True):
+    def get_equations_for(self, symbols, sort_by_input_symbols=False, excluded_symbols=[]):
         """Get all equations for given collection of symbols
 
         Results are sorted in topographical order.
         :param symbols: the symbols to get the equations for
-        :param lexicographical_sort: indicates whether the result is sorted in lexicographical order first
+        :param sort_by_input_symbols: indicates whether the sorting in symbols determines the sorting of the output
```
jonc125: Doing this manually just feels wrong. Surely there's something in the NetworkX library we can get to do this for us?

For instance, the [lexicographic](https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.lexicographical_topological_sort.html) sort takes a `key` argument to resolve ambiguities in the sort order. Would that work?

cellmlmanip/model.py
MauriceHendrix: Not really if it's false we still check the required_symbols, we're going to need those one way or another.
```        for symbol in to_search_in:
            # Ignore symbols we don't need
            if symbol not in required_symbols:
                continue```

cellmlmanip/model.py
MauriceHendrix: What you'd have to do is write a key function key such that 
given input [x1...xN] to the get_equations_for fuction
key(x1) < ... < ... key(xN) (which is straightforward) but also
key (y1) < ... < key(yN) where yn is a parent of xn

I don't really see any way of doing this without actually doing the sort unless I'm missing something?

jonc125:
The purpose of this sorting is purely to make it easier to compare with pycml output right? So we're getting rid of it and reverting to lexicographical sort once we can generate new reference data? In which case it's probably not worth worrying about!

--
Written on a smartphone; please excuse brevity!

________________________________

MauriceHendrix:
> The purpose of this sorting is purely to make it easier to compare with pycml output right? So we're getting rid of it and reverting to lexicographical sort once we can generate new reference data? In which case it's probably not worth worrying about!

Yes I was going to contact you about that actually. in fact for now maybe don't even merge this in yet. Once I have the normal model properly working I'll write a test that can cope with equations not being in the same order and then test if we actually really need this or not.

jonc125:
Sounds good!

--
Written on a smartphone; please excuse brevity!

________________________________

MauriceHendrix:
Right so it looks like this isn't necessary at all as it looks like I get things back in a suitable order (even if not the same as pycml)

PR #88, opened 2019-09-30
added dimensionally_equivalent

MauriceHendrix:
added dimensionally_equivalent to allow checking if 2 symbols are dimensionally equivalent

### DIFF
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -250,7 +250,7 @@ class UnitStore(object):
                                        self.summarise_units(symbol2))
             return True
         except pint.errors.DimensionalityError:
-            return False        
+            return False
 
 
 class UnitCalculator(object):
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -244,7 +244,7 @@ class UnitStore(object):
         return self.convert_to(quantity, to_unit).magnitude
 
     def dimensionally_equivalent(self, symbol1, symbol2):
-        """Returns whether symbol1 and symbol2 are dimensionally_equivalent (same units ignoring a calling factor)"""
+        """Returns whether symbol1 and symbol2 are dimensionally_equivalent (same units ignogging a calling factor)"""
         try:
             self.get_conversion_factor(1 * self.summarise_units(symbol1),
                                        self.summarise_units(symbol2))
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -243,6 +243,15 @@ class UnitStore(object):
         """Returns the magnitude multiplier required to convert from_unit to to_unit """
         return self.convert_to(quantity, to_unit).magnitude
 
+    def dimensionally_equivalent(self, symbol1, symbol2):
+        """Returns whether symbol1 and symbol2 are dimensionally_equivalent (same units ignoging a caling factor)"""
+        try:
+            self.get_conversion_factor(1 * self.summarise_units(symbol1),
+                                       self.summarise_units(symbol2))
+            return True
+        except pint.errors.DimensionalityError:
+            return False        
+
 
 class UnitCalculator(object):
     """Evaluates a Sympy expression to determine its units. Note: only supports subset of Sympy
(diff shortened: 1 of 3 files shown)


cellmlmanip/units.py
```diff
@@ -243,6 +243,15 @@ def get_conversion_factor(self, quantity, to_unit):
         """Returns the magnitude multiplier required to convert from_unit to to_unit """
         return self.convert_to(quantity, to_unit).magnitude
 
+    def dimensionally_equivalent(self, symbol1, symbol2):
+        """Returns whether symbol1 and symbol2 are dimensionally_equivalent (same units ignoging a caling factor)"""
```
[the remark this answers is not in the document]
MauriceHendrix: fixed

cellmlmanip/units.py
```diff
@@ -243,6 +243,15 @@ def get_conversion_factor(self, quantity, to_unit):
         """Returns the magnitude multiplier required to convert from_unit to to_unit """
         return self.convert_to(quantity, to_unit).magnitude
 
+    def dimensionally_equivalent(self, symbol1, symbol2):
+        """Returns whether symbol1 and symbol2 are dimensionally_equivalent (same units ignoring a calling factor)"""
+        try:
+            self.get_conversion_factor(1 * self.summarise_units(symbol1),
```
MichaelClerx: Why `1 * ` ???

tests/test_hodgkin.py
```diff
@@ -307,3 +307,10 @@ def test_get_equations_for(self, graph, model):
         # Check not lexicographical sorted equations against expected equations
         for i in range(len(unordered_equations)):
             assert str(unordered_equations[i]) == str(unordered_ref_eq[i])
+
+    def test_dimensionally_equivalent(self, model):
```
MichaelClerx: This doesn't go here! How would I ever know that `test_hodgkin` was the place to look for a test of a method in the UnitStore class?

There should be a test_unit_store somewhere, or perhaps a test_units, where this will fit much better

MauriceHendrix: Made the requested change.
However if you do have a point but if you follow that through the following tests shouldn't be there either as they are testing the respective functions and not the model.

`test_get_symbol_by_ontology_term`
`test_get_ontology_terms_by_symbol`
`test_has_ontology_annotation`
`test_get_equations_for`

Of course that would need a separate pull request!

MichaelClerx: Good point!
Thanks for moving it

cellmlmanip/units.py
MauriceHendrix: because `def get_conversion_factor(self, quantity, to_unit):`
`self.summarise_units(symbol1)`  is a unit and `1 * self.summarise_units(symbol1)` is a quanity

MichaelClerx: Thanks!

tests/test_hodgkin.py
MichaelClerx: This bit can go then?

MauriceHendrix: forgot to commit sorry :)

tests/test_units.py
```diff
@@ -144,3 +148,16 @@ def test_expression_printer(self, quantity_store):
         assert printer.doprint(a / b) == 'a[meter]/b[second]'
         assert printer.doprint(_2 * y) == '2.000000[dimensionless]*y[volt]'
         assert printer.doprint(sp.Derivative(a, b)) == 'Derivative(a[meter], b[second])'
+
+    def test_dimensionally_equivalent(self):
+        hodgkin_cellml = os.path.join(
```
MichaelClerx: Is this our go-to test model? If so, perhaps we should have a pytest fixture for loading it

MichaelClerx: https://github.com/ModellingWebLab/cellmlmanip/blob/develop/tests/test_parser.py#L10-L22

MichaelClerx: I think this file has variables specifically for testing unit conversion: https://github.com/ModellingWebLab/cellmlmanip/blob/develop/tests/cellml_files/test_simple_odes.cellml

MauriceHendrix: Not sure right now there is only 1 test that uses it in there, but sure a fixture may be useful

jonc125: A shared fixture in `conftest.py` might make most sense?

PR #89, opened 2019-09-30
Changed get_symbols to accept a collection (list) of expressions

MauriceHendrix:
Changed get_symbols to accept a collection (list) of expressions.
To bring it in line with get_equations_for

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -600,7 +600,7 @@ class Model(object):
                            variable.dummy, variable.type, variable_type)
 
     def get_symbols(self, expressions):
-        """Returns the symbols in a collectiion of expressions. Works differently to Sympy atoms() because a
+        """Returns the symbols in a collection of expressions. Works differently to Sympy atoms() because a
         Derivative instance is regarded as a single symbol"""
         symbols = set()
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -309,7 +309,7 @@ class Model(object):
             equation_count += 1
 
             # Determine LHS. We should only every have one symbol (or derivative)
-            lhs_symbol = self.get_symbols([equation.lhs])
+            lhs_symbol = self.get_symbols_for([equation.lhs])
             assert len(lhs_symbol) == 1
             lhs_symbol = lhs_symbol.pop()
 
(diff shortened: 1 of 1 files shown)


cellmlmanip/model.py
```diff
@@ -599,18 +599,19 @@ def _set_variable_type(variable, variable_type):
             logger.warning('Variable %s already has type=="%s". Skip setting "%s"',
                            variable.dummy, variable.type, variable_type)
 
-    def get_symbols(self, expr):
-        """Returns the symbols in an expression. Works differently to Sympy atoms() because a
+    def get_symbols(self, expressions):
+        """Returns the symbols in a collectiion of expressions. Works differently to Sympy atoms() because a
```
skeating: spelling collection

cellmlmanip/model.py
```diff
@@ -599,18 +599,19 @@ def _set_variable_type(variable, variable_type):
             logger.warning('Variable %s already has type=="%s". Skip setting "%s"',
                            variable.dummy, variable.type, variable_type)
 
-    def get_symbols(self, expr):
-        """Returns the symbols in an expression. Works differently to Sympy atoms() because a
+    def get_symbols(self, expressions):
+        """Returns the symbols in a collection of expressions. Works differently to Sympy atoms() because a
```
MichaelClerx: Please rewrite this a bit clearer. I thought it was saying that Sympy's `atoms()` regarded a derivative as a single symbol, but it seems from the code that you mean the opposite.

MichaelClerx: To be honest, don't mention Sympy at all (we're not documenting Sympy, plus what if they change their code?)
Just add a note saying that derivatives count as a single symbol.

cellmlmanip/model.py
```diff
@@ -599,18 +599,19 @@ def _set_variable_type(variable, variable_type):
...
         Derivative instance is regarded as a single symbol"""
         symbols = set()
```
MichaelClerx: Second issue with this method: A sympy derivative is not a symbol

https://github.com/sympy/sympy/blob/master/sympy/core/function.py#L1024

It's an expression!

cellmlmanip/model.py
```diff
@@ -599,18 +599,19 @@ def _set_variable_type(variable, variable_type):
             logger.warning('Variable %s already has type=="%s". Skip setting "%s"',
                            variable.dummy, variable.type, variable_type)
 
-    def get_symbols(self, expr):
-        """Returns the symbols in an expression. Works differently to Sympy atoms() because a
+    def get_symbols(self, expressions):
```
MichaelClerx: Should probably be called something like `get_symbols_for` or `get_symbols_for_expression`. I'd expect `model.get_symbols()` to return the symbols in the model

cellmlmanip/model.py
MauriceHendrix: I just changed what was changed but I don't mind updating the doc string

cellmlmanip/model.py
MauriceHendrix: Updated the comment to reflect this

cellmlmanip/model.py
MauriceHendrix: Makes sense, done

PR #98, opened 2019-10-02
Shortcut conversion factor

skeating:


cellmlmanip/units.py
```diff
@@ -338,24 +339,52 @@ def summarise_units(self, expr: sympy.Expr):
...
-    def get_conversion_factor(self, quantity, to_unit):
-        """Returns the magnitude multiplier required to convert from_unit to to_unit
+    def get_conversion_factor(self,
+                              to_unit,
+                              from_unit=None,
+                              quantity=None,
+                              expression=None):
+        """Returns the magnitude multiplier required to convert a unit to the specified unit.
+
+        Note this will work on either a unit, a quantity or an expression. If more than one
```
MauriceHendrix: Hang on,
this is inconsistent. as later on you use an assert to say that at least 2 of those should be None. 
Personally I think that makes more sense than accepting arguments that you're not using, but either way the description should be consistent with what's actually happening.

skeating: Good catch - thanks

tests/test_unit_conversion.py
```diff
@@ -20,3 +25,94 @@ def test_add_preferred_custom_unit_name(model):
...
+def test_conversion_factor_bad_types(simple_model):
+    simple_model.get_equation_graph(True)  # set up the graph - it is not automatic
+    symbol_b1 = simple_model.get_symbol_by_cmeta_id("b_1")
+    equation = simple_model.get_equations_for([symbol_b1])
+    expression = equation[0].lhs
+    to_unit = simple_model.units.ureg('us').units
+    from_unit = simple_model.units.summarise_units(expression)
+    quantity = 1 * from_unit
+    # no source unit
+    try:
```
jonc125: These won't fail the test if no exception is raised! You should use something like
```python
with pytest.raises(AssertionError) as excinfo:
    simple_model.units.get_conversion_factor(to_unit=to_unit)
assert 'No unit...' in excinfo.value
```
or
```python
with pytest.raises(AssertionError, match='^No unit given.*'):
    simple_model.units.get_conversion_factor(to_unit=to_unit)
```

tests/test_unit_conversion.py
```diff
@@ -20,3 +25,94 @@ def test_add_preferred_custom_unit_name(model):
...
+    from_unit = simple_model.units.summarise_units(expression)
+    quantity = 1 * from_unit
+    # quantity to unit
+    assert simple_model.units.get_conversion_factor(to_unit=to_unit, quantity=quantity) == 1
+    # unit to unit
+    assert simple_model.units.get_conversion_factor(to_unit=to_unit, from_unit=from_unit) == 1
+    # expression to unit
+    assert simple_model.units.get_conversion_factor(to_unit=to_unit, expression=expression) == 1
+
+def test_conversion_factor(simple_model):
```
jonc125: This appears to be identical to the test above?

skeating: Indeed

MichaelClerx:
Hey @MauriceHendrix , do you approve of the changes?

MauriceHendrix:
> Hey @MauriceHendrix , do you approve of the changes?

looks good to me. I'll need to update some code on my end but shouldn't be a big deal. Has it been marged into the master yet?

skeating:
@MauriceHendrix You need to confirm the changes/dismiss the review when you asked for changes before I can merge

MauriceHendrix:
> @MauriceHendrix You need to confirm the changes/dismiss the review when you asked for changes before I can merge

Aah ok I didn't realise

PR #100, opened 2019-10-02
replicate the bug fixes #92

### LINKED ISSUE #92: get_initial_value bug and possible get_equations_for bug
I think I have discovered a bug (let me know if I'm missing something!) 
This bug is preventing me from continuing with chaste code gen!
The model which I'm working with is attached

What I'm doing is:
`membrane_capacitance = self._model.get_symbol_by_ontology_term(self._OXMETA, "membrane_capacitance")`
`initial_value = self._model.get_initial_value(membrane_capacitance))`
and I get a `KeyError`

The model has a membrane_capacitance with an initial value.

Second issue, but I am less sure of this: should the capacitence not show up in the following? (it doesn't):
`membrane_stimulus_current = self._model.get_symbol_by_ontology_term(self._OXMETA, "membrane_stimulus_current")`
`self._model.get_equations_for([membrane_stimulus_current ])`

[aslanidi_model_2009.zip](https://github.com/ModellingWebLab/cellmlmanip/files/3676516/aslanidi_model_2009.zip)

skeating:
Looking into bug https://github.com/ModellingWebLab/cellmlmanip/issues/92
Fixes #92


cellmlmanip/model.py
```diff
@@ -643,7 +643,13 @@ def get_initial_value(self, symbol):
         :param symbol: Sympy Dummy object of required symbol
         :return: float of initial value
         """
-        return float(self.graph.nodes[symbol]['initial_value'])
+        # fixed for issue where graph.nodes[symbol] does not have an 'initial_value' entry
```
jonc125: s/fixed/Fix/

cellmlmanip/model.py
```diff
@@ -643,7 +643,13 @@ def get_initial_value(self, symbol):
...
+        node = self.graph.nodes[symbol]
+        if 'initial_value' in node:
+            return float(node['initial_value'])
+        # no initial_value entry try the dummy_metadata
```
jonc125: Would it be better just to do this approach always? Then we wouldn't need any comments here, just the two lines of code below!

tests/test_aslanidi_model.py
```diff
@@ -0,0 +1,31 @@
...
+            os.path.dirname(__file__),
+            "cellml_files",
+            "aslanidi_model_2009.cellml"
+        )
+        return load_model(cellml)
+
+    def test_initial_value_capacitance(self, model):
+        model.get_equation_graph()
+        membrane_capacitance = model.get_symbol_by_ontology_term(OXMETA, "membrane_capacitance")
+        # was raising KeyError but should not
```
jonc125: This commented out bit should be removed.

tests/test_unit_conversion.py
```diff
@@ -89,4 +89,3 @@ def test_conversion_factor_same_units(simple_model):
     assert simple_model.units.get_conversion_factor(to_unit=to_unit, from_unit=from_unit) == 1
     # expression to unit
     assert simple_model.units.get_conversion_factor(to_unit=to_unit, expression=expression) == 1
-
```
jonc125: Isn't it best practice to keep a newline at the end of source files?

cellmlmanip/model.py
skeating: My rationale was that the original seemed to work for most things but I've simplified it until we hit a case where that one doesnt work :-)

skeating:
@MauriceHendrix since you reported the bug it would be great if you check this fixes your issue

MauriceHendrix:
> @MauriceHendrix since you reported the bug it would be great if you check this fixes your issue

Works great thanks!

Issue #102, opened 2019-10-17
Cellmlmanip breaks when using newer networkx


MichaelClerx:
Currently pinned to 2.3, breaks if using 2.4

skeating:
Yes I hit this using weblab-fc. Do we want to fix it by insisting that we have 2.3 or fix for 2.4

MichaelClerx:
Someone will have to fix it so that it works for both

MichaelClerx:
Or possibly only 2.4 :D

MichaelClerx:
https://networkx.github.io/documentation/stable/release/release_2.4.html

Think .node was deprecated since 2.1, if I'm reading this correctly. So could perhaps require 2.1 or greater and apply this fix?

MichaelClerx:
```G.node –> use G.nodes```

PR #103, opened 2019-10-17
Make cellmlmanip compatible with Python 3.5

MichaelClerx:
It's the default in ubuntu LTS 16 which has support until 2021, and 3.5 (and older!) will be used in university systems for long after.

In addition, checking on 3.5 I found a couple of instances where we relied on Python3.6 dict ordering, which we should never do, as this is an implementation artefact.

### DIFF
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -15,9 +15,10 @@ class Transpiler(object):
     """
     Handles conversion of MathmL to Sympy exprerssions.
 
-    :param dummify: A bool that ???
-    :param symbol_prefix: A str that ???
-    :param symbol_lookup: A dict mapping ..?
+    :param dummify: Set to ``True`` to create dummy symbols for variables instead of sympy Symbols
+    :param symbol_prefix: An optional prefix to add to all symbols
+    :param symbol_lookup: A dict mapping variable names (including the optional ``symbol_prefix``) to
+        predefined Symbol or Dummy objects
     """
 
     def __init__(self, dummify=False, symbol_prefix=None, symbol_lookup=dict()):
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -10,7 +10,6 @@ import sympy
 
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
 
 
 class Transpiler(object):
--- a/cellmlmanip/mathml2sympy/transpiler.py
+++ b/cellmlmanip/mathml2sympy/transpiler.py
@@ -3,7 +3,6 @@
 Content Markup specification: https://www.w3.org/TR/MathML2/chapter4.html
 """
 import logging
-from typing import Dict, List
 from xml.dom import Node, minidom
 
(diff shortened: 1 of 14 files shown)


MichaelClerx:
Appveyor is failing due to #102

jonc125:
Ordered dict may have been an implementation detail in 3.6, but it has become an official part of the language spec in later versions!

MichaelClerx:
> Ordered dict may have been an implementation detail in 3.6, but it has become an official part of the language spec in later versions!

Good! But doesn't change much for older versions

cellmlmanip/model.py
```diff
@@ -74,39 +75,59 @@ def __str__(self) -> str:
...
+        :param equation: A ``sympy.Eq`` object.
+        """
         assert isinstance(equation, sympy.Eq), 'Equation expression must be equality'
         self.equations.append(equation)
 
-    def add_number(self, *,
-                   number: sympy.Number, units: str, dummy: sympy.Dummy = None) -> sympy.Dummy:
-        """Add metadata about a dummy symbol that represents a number in equations. Returns the
-        sympy.Dummy object used to represent this number"""
+    # TODO: Do we need the * here?
```
jonc125: Yep, says arguments after must be keyword-only. Same below.

MichaelClerx: What's the rationale for that?

MichaelClerx:
> Ordered dict may have been an implementation detail in 3.6, but it has become an official part of the language spec in later versions!

A more subtle issue is that a lot of the ordering stems from NetworkX objects, which apparently use a plain python dict so that the behaviour changes with python 3.6. But the NetworkX docs still say the ordering is non-deterministic

tests/test_state_symbols.py
```diff
@@ -27,21 +28,6 @@ def test_state_symbols(self, model):
         state_symbols = model.get_state_symbols()
         state_symbols_ordered_by_input = model.get_state_symbols(order_by_order_added=True)
         assert len(state_symbols) == len(state_symbols_ordered_by_input) == 29
-        assert str(state_symbols) == \
```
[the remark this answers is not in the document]
jonc125: No, the line you removed.

cellmlmanip/model.py
jonc125: It's a defensive programming safety feature. Means users can't get the order of arguments wrong, and makes user code more readable.

cellmlmanip/mathml2sympy/transpiler.py
```diff
@@ -3,31 +3,32 @@
...
 class Transpiler(object):
-    """Transpiler class handles conversion of MathmL to Sympy exprerssions"""
+    """
+    Handles conversion of MathmL to Sympy exprerssions.
 
-    def __init__(self,
-                 dummify: bool = False,
-                 symbol_prefix: str = None,
-                 symbol_lookup=dict()) -> None:
+    :param dummify: A bool that ???
```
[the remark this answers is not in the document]
jonc125: Because some bits were written a while ago when Asif was still figuring out CellML, and concentrating on getting an initial working 'tracer bullet' to achieve Web Lab project goals, rather than writing libCellML in Python...

MichaelClerx:
Made the suggested changes!

PR #104, opened 2019-10-17
Replaced graph.node by graph.nodes (deprecated since networkx 2.1).

MichaelClerx:
See #102


MichaelClerx:
This works now if you could have a quick look @skeating !

Issue #101, opened 2019-10-17
Why is coverage config file not in root?


MichaelClerx:
https://github.com/ModellingWebLab/cellmlmanip/pull/66/files

stuck it in `cellmlmanip`. Is there any reason for that?

jonc125:
Copying a previous project I think; no particular reason.

jonc125:
See also #76. Fixed by #105.

PR #105, opened 2019-10-17
Moved coveragerc so all infra config files are in the root, not in th…

MichaelClerx:
…e python module.

See #101


skeating:
Does something need to change in the travis configuration so that that works ?

MichaelClerx:
Woops! Yes just fixing it now :-D

PR #109, opened 2019-10-24
added option to be able to not recurse down the equation graph for get_equations_for

MauriceHendrix:


jonc125:
Presumably you'll call this repeatedly to do BFS when finding i_ionic?

MauriceHendrix:
> Presumably you'll call this repeatedly to do BFS when finding i_ionic?
Nope just the top level equations. (those are the terms that end up in `const double var_chaste_interface__i_ionic =`

jonc125:
> How would I be able to tell? I can't see the difference between that and what I had previously ?

That's why pycml keeps recursing one level down until it finds one or more variables with units matching `i_stim`.

MichaelClerx:
If it gets quite specific, maybe it should go in weblab_cg instead of cellmlmanip?
Can see a case either way :-)

jonc125:
Yeah, I think the method as-is is fine for cellmlmanip; the extra cleverness would go in the Chaste code gen.

PR #112, opened 2019-11-04
Provided feature to allow ordering derivative symbols in document order.

MauriceHendrix:
Needed for chaste codegen


cellmlmanip/model.py
```diff
@@ -495,15 +495,24 @@ def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True):
 
         return eqs
 
-    def get_derivative_symbols(self):
+    def get_derivative_symbols(self, order_by_order_added=False):
         """Returns a list of derivative symbols found in the given model graph.
+
+        :param order_by_order_added: indicates whether state_symbols are sorted in the order they appear in the model.
+        Please Note: the order of returned derivative symbols is not guaranteed
```
MichaelClerx: Is this part of the :param: block? Could you change the sentence to put "without the flag" at the start of the sentence and remove the "on Python <= 3.5" block? The contract for the method should be the same regardless of the python version

MichaelClerx: Also, if the options are (1) a particular order, or (2) no particular order, then why not always do (1) ??
So just drop the param and always sort

tests/test_derivative_symbols.py
```diff
@@ -0,0 +1,60 @@
+import os
```
MichaelClerx: Why is this a separate file? Would expect this in test_model.py or something that relates to the class name

MauriceHendrix: because yo complained when i put the test for state_symbols in the model and told me to make a seperate est. this is the same thing so also a separete test for consistancy!

cellmlmanip/model.py
MauriceHendrix: It kind of is part of the param block in a way as otherwise you'd have to refer to the param before it's explained.

I don't mind always sorting, but it's something we should discuss.  I didn't want to change default behavior if I don't have to and also sorting makes it slower in theory (though it's not really noticeable)

MichaelClerx: Yeah I don't think the sorting will slow us down much...

tests/test_derivative_symbols.py
MichaelClerx: Can we at least call it something with "model" in the filename? Need to be able to find these tests later

MauriceHendrix: Hhm then it would be better to move it into the model. I don't mind that if others don't mind.

cellmlmanip/model.py
MauriceHendrix: What I could to is make the flag default True then if it becomes performance critical there is the option to turn it off.

tests/test_derivative_symbols.py
MichaelClerx: I wouldn't mind that. But if you say we've discussed it before.. I can't find that discussion though!

MichaelClerx: Probably good to see what reasons / arguments / conclusions we had :D

MauriceHendrix: I can't find it sorry

cellmlmanip/model.py
```diff
@@ -497,18 +497,17 @@ def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True):
...
         """
-        return [v for v in self.graph if isinstance(v, sympy.Derivative)]
+        derivative_symbols = [v for v in self.graph if isinstance(v, sympy.Derivative)]
+        return sorted(derivative_symbols, key=lambda state_var: self.get_meta_dummy(state_var.args[0]).order_added)
 
     def get_state_symbols(self, order_by_order_added=False):
         """Returns a list of state variables found in the given model graph.
-
-        :param order_by_order_added: indicates whether state_symbols are sorted in the order they appear in the model.
+           The list is ordered by appearance in the cellml document.
```
MichaelClerx: Indenting!

cellmlmanip/model.py
```diff
@@ -497,18 +497,17 @@ def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True):
 
     def get_derivative_symbols(self):
         """Returns a list of derivative symbols found in the given model graph.
+           The list is ordered by appearance in the cellml document.
         """
-        return [v for v in self.graph if isinstance(v, sympy.Derivative)]
+        derivative_symbols = [v for v in self.graph if isinstance(v, sympy.Derivative)]
+        return sorted(derivative_symbols, key=lambda state_var: self.get_meta_dummy(state_var.args[0]).order_added)
 
     def get_state_symbols(self, order_by_order_added=False):
```
MichaelClerx: The argument is still there :D

tests/test_derivative_symbols.py
MauriceHendrix: I can't find it sorry

cellmlmanip/model.py
MauriceHendrix: uups sorry, fixed now!

cellmlmanip/model.py

tests/test_aslanidi_model.py
```diff
@@ -27,3 +27,64 @@ def test_initial_value_voltage(self, model):
...
+
+    def test_state_symbols(self, model):
+        """
+        _membrane$V
+        _sodium_channel_m_gate$m
+        _sodium_channel_h_gate$h
+        _potassium_channel_n_gate$n
+        """
+        state_symbols = model.get_state_symbols()
+        state_symbols_ordered_by_input = model.get_state_symbols(order_by_order_added=True)
```
MichaelClerx: This bit needs updating

MauriceHendrix: o yes good spot, done

PR #117, opened 2019-11-14
New method to get rdf annotation more generically

MauriceHendrix:


cellmlmanip/model.py
```diff
@@ -555,21 +555,57 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
         """
-        # Convert property and value to RDF nodes
-        # TODO: Eventually a different form for predicate and object may be
-        #       accepted.
         assert len(predicate) == 2
         predicate = create_rdf_node(*predicate)
         if object_ is not None:
-            assert len(object_) == 2
-            object_ = create_rdf_node(*object_)
+            if isinstance(object_, str):
```
jonc125: I'd prefer to merge this logic into `create_rdf_node` (I think there's some commented code in there for how it used to work in pycml!)

cellmlmanip/model.py
```diff
@@ -555,21 +555,57 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
             raise ValueError('Multiple variables annotated with {%s}%s' %
                              (namespace_uri, local_name))
 
-    def _get_symbols_by_rdf(self, predicate, object_=None):
+    def get_rdf_annotation(self, predicate):
+        """Searches the RDF graph for variables annotated with the given predicate
```
jonc125: Except this doesn't just find variables, judging by the example given!

Also, how do you distinguish in the results what has been found as the subject, between variables and non-variables? I think you probably know generally whether you're hunting for model annotations or variable annotations; in the former case you want to specify the subject in your query for efficiency, and in the latter you want to ensure you only get variables.

MauriceHendrix: That's correct it doesn't just find variables, will update the docstring. It's specifically for model annotations. But it could be used for any rdf. There is absolutly no link back to model variables, if you want that use get_symbols_by_rdf.

   <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" >
      <rdf:Description rdf:about="#test_simple_odes">
          <named-attribute xmlns="https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#">
             <rdf:Description>
                <name rdf:datatype="http://www.w3.org/2000/10/XMLSchema#string">SuggestedForwardEulerTimestep</name>
                <value rdf:datatype="http://www.w3.org/2000/10/XMLSchema#double">0.0002</value>
             </rdf:Description>
          </named-attribute>
      </rdf:Description>
   </rdf:RDF>

cellmlmanip/model.py
MauriceHendrix: done

cellmlmanip/model.py
```diff
@@ -555,21 +555,51 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+        """Searches the RDF graph for rdf annotation with the given predicate
+        and (e.g. "pycml:named-attribute) and returns the associated annotation.
+
+        ``predicate`` must be a ``(namespace, local_name)`` tuple.
+        The method will return a dict following the same structure as the rdf annotation.
+        e.g. when looking for named-attribute (of which there could be multiple):
+        "{'model_name': {'named-attribute': [{'name': 'SuggestedForwardEulerTimestep', 'value': '0.002'}]}}"
+        Please note: for getting annotated model variables directly, use get_symbols_by_rdf
+        """
+        def _get_annotation(tuples):
```
jonc125: Call the argument `triples` for clarity.

cellmlmanip/model.py
```diff
@@ -555,21 +555,51 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
             raise ValueError('Multiple variables annotated with {%s}%s' %
                              (namespace_uri, local_name))
 
-    def _get_symbols_by_rdf(self, predicate, object_=None):
+    def get_rdf_annotation(self, predicate):
```
jonc125: If you're only using this on model annotations, I'd add a subject to the original call. Even if it can be used more broadly, this can be optional. Perhaps call it `get_rdf_annotations(self, subject=None, predicate=None, object_=None)`?

There's potential for this to confuse if two subjects have the same local name; the docs should at least warn of this. Using full URIs would be safer though (and is what pycml does, so I'd rather we didn't regress!).

I'm not sure converting the RDF structure to a dict helps much, actually; to me it just seems to add complexity. Could this not just return the matching triples, and let the user decide what to do with them?

cellmlmanip/model.py
```diff
@@ -555,21 +555,51 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+        The method will return a dict following the same structure as the rdf annotation.
+        e.g. when looking for named-attribute (of which there could be multiple):
+        "{'model_name': {'named-attribute': [{'name': 'SuggestedForwardEulerTimestep', 'value': '0.002'}]}}"
+        Please note: for getting annotated model variables directly, use get_symbols_by_rdf
+        """
+        def _get_annotation(tuples):
+            """function that populates the annotation dict. It unpacks objects that are themselves rdf triples"""
+            annotation_dict = dict()
+            subject_name = None
+            for result in tuples:
```
jonc125: I'd make this `for subj, pred, obj in triples:` so it's clearer what `result[0]` etc are below.

cellmlmanip/model.py
```diff
@@ -555,21 +555,51 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+                predicate_name = str(result[1])[str(result[1]).rfind('#') + 1:]
+                if isinstance(result[0], rdflib.URIRef):
+                    subject_name = str(result[0].lstrip('#'))
+                    if subject_name not in annotation_dict:
+                        annotation_dict[subject_name] = dict()
+                if isinstance(result[2], rdflib.Literal):
+                    annotation_dict[predicate_name] = str(result[2]).strip()
```
jonc125: You're making assumptions about the structure of the annotations here! It'll break in general.

cellmlmanip/model.py
```diff
@@ -555,21 +555,51 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+        # Find symbols
+        rdf_annotation = _get_annotation(list(self.rdf.triples((None, predicate, None))))
+
+        return rdf_annotation
+
+    def get_symbols_by_rdf(self, predicate, object_=None):
         """Searches the RDF graph for variables annotated with the given predicate
         and object (e.g. "is oxmeta:time") and returns the associated symbols.
 
         Both ``predicate`` and ``object_`` (if given) must be
```
jonc125: This doc is no longer true.

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,20 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
 '''
 
 
-def create_rdf_node(namespace_uri, local_name):
+def create_rdf_node(namespace_uri, local_name=None):
```
jonc125: This is horrible! It should take a single argument always, since that's what it's always called with now. i.e. `def create_rdf_node(node_content):`

The docs also need updating.

Then the body can be like
```python
if isinstance(node_content, str):
    return rdflib.Literal(node_content)
else:
    namespace_uri, local_name = node_content
    # continue as originally
```

cellmlmanip/model.py
MauriceHendrix: fair point

cellmlmanip/model.py
MauriceHendrix: nice idea why didn't I think of that :)

cellmlmanip/model.py

cellmlmanip/rdf.py
MauriceHendrix: hhm good point. I was struggeling to keep the same method signature but maybe that's not possible

cellmlmanip/model.py
MauriceHendrix: It's true that it's a little complex and it does wrap annotation in a structure that still needs to be processed in general (although in my case I get exactly what I need).

Reason I did this was that returning the triples doesn't give me the annotation I have asked for and I'd have to start following the rdf graph to get to it. If we want this that's fine.


As an example I get the following triple:

`(rdflib.term.URIRef('#bondarenko_model_2004_apex'), rdflib.term.URIRef('https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#named-attribute'), rdflib.term.BNode('N9d040fafd83243b4abbb9275d4eb8eba'))`

For this RDF:

```
   <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" >
      <rdf:Description rdf:about="#test_simple_odes">
          <named-attribute xmlns="https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#">
             <rdf:Description>
                <name rdf:datatype="http://www.w3.org/2000/10/XMLSchema#string">SuggestedForwardEulerTimestep</name>
                <value rdf:datatype="http://www.w3.org/2000/10/XMLSchema#double">0.0002</value>
             </rdf:Description>
          </named-attribute>
   </rdf:RDF>
```

jonc125:
So, my thought was if we enhanced `create_rdf_node` to do everything it did in pycml (with tests :) ) then we could make `get_rdf_annotations()` essentially just a facade to `self.rdf.triples`, i.e.
```python
def get_rdf_annotations(self, subject=None, predicate=None, object_=None):
    if not subject is None:
        subject = create_rdf_node(subject)
    # same with predicate, object_
    return self.rdf.triples((subject, predicate, object_))
```

Another useful helper function might be:
```python
def get_rdf_value(self, subject, predicate):
    """Get the value of an RDF object connected to ``subject`` by ``predicate``."""
    triples = self.get_rdf_annotations(subject, predicate)
    assert len(triples) == 1
    value = str(triples[0][2]).strip()  # Could make this cleverer by considering data type if desired
    return value
```

We'd also need to record the model's cmeta_id, perhaps as `self.cmeta_id` on the `Model` class. And have `Model.__init__` set something like `self.rdf_identity = create_rdf_node(fragment_id=self.cmeta_id)`.

Then your calling code looking for the named attributes could be something like:
```python
pycml_ns = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
named_attrs = model.get_rdf_annotations(subject=model.rdf_identity, predicate=(pycml_ns, 'named-attribute'))
for _, _, attr in named_attrs:
    name = model.get_rdf_value(subject=attr, predicate=(pycml_ns, 'name'))
    value = model.get_rdf_value(subject=attr, predicate=(pycml_ns, 'value'))
    # Do stuff with name & value...
```

MauriceHendrix:
> So, my thought was if we enhanced `create_rdf_node` to do everything it did in pycml (with tests :) ) then we could make `get_rdf_annotations()` essentially just a facade to `self.rdf.triples`, i.e.
> 
> ```python
> def get_rdf_annotations(self, subject=None, predicate=None, object_=None):
>     if not subject is None:
>         subject = create_rdf_node(subject)
>     # same with predicate, object_
>     return self.rdf.triples((subject, predicate, object_))
> ```
> 
> Another useful helper function might be:
> 
> ```python
> def get_rdf_value(self, subject, predicate):
>     """Get the value of an RDF object connected to ``subject`` by ``predicate``."""
>     triples = self.get_rdf_annotations(subject, predicate)
>     assert len(triples) == 1
>     value = str(triples[0][2]).strip()  # Could make this cleverer by considering data type if desired
>     return value
> ```
> 
> We'd also need to record the model's cmeta_id, perhaps as `self.cmeta_id` on the `Model` class. And have `Model.__init__` set something like `self.rdf_identity = create_rdf_node(fragment_id=self.cmeta_id)`.
> 
> Then your calling code looking for the named attributes could be something like:
> 
> ```python
> pycml_ns = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
> named_attrs = model.get_rdf_annotations(subject=model.rdf_identity, predicate=(pycml_ns, 'named-attribute'))
> for _, _, attr in named_attrs:
>     name = model.get_rdf_value(subject=attr, predicate=(pycml_ns, 'name'))
>     value = model.get_rdf_value(subject=attr, predicate=(pycml_ns, 'value'))
>     # Do stuff with name & value...
> ```
Hhm model.rdf_identity doesn't exist and I don't know how to make a valid uriref out of the model name

jonc125:
Yeah, `model.cmeta_id` doesn't exist either, it'll need to be added during the model parsing process. I don't think cellmlmanip stores the model's `cmeta:id` attribute anywhere at present.

cellmlmanip/model.py
```diff
@@ -78,9 +78,11 @@ class Model(object):
 
     :param name: the name of the model e.g. from <model name="">
     """
-    def __init__(self, name):
+    def __init__(self, name, cmeta_id):
```
jonc125: Need to update docstring above too.

MauriceHendrix: o yeah

cellmlmanip/model.py
```diff
@@ -77,10 +77,13 @@ class Model(object):
...
 
     :param name: the name of the model e.g. from <model name="">
+    :param cmeta_id: the cmeta_id of the model e.g. from <model cmeta_id="">
     """
-    def __init__(self, name):
+    def __init__(self, name, cmeta_id):
 
         self.name = name
+        self.cmeta_id = cmeta_id
+        self.rdf_identity = rdflib.URIRef('#' + cmeta_id) if cmeta_id is not None else None
```
jonc125: I'd just say `if cmeta_id else` for simplicity! Also, you could use your feature in `create_rdf_node`:
```python
   self.rdf_identity = create_rdf_node('#' + cmeta_id) if cmeta_id else None
```

cellmlmanip/model.py
```diff
@@ -555,21 +558,29 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
             raise ValueError('Multiple variables annotated with {%s}%s' %
                              (namespace_uri, local_name))
 
-    def _get_symbols_by_rdf(self, predicate, object_=None):
+    def get_rdf_annotations(self, subject=None, predicate=None, object=None):
```
jonc125: Use `object_` rather than `object` to avoid shadowing the built-in.

cellmlmanip/model.py
```diff
@@ -555,21 +558,29 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+        """Searches the RDF graph and returns the triples for the given (subject, predicate, object)"""
```
jonc125: Can we expand this docstring to be more informative? So the one-liner could end 'triples matching the given parameters', then blank line, then more usage notes. Cover things like if a param is `None` then any triple matches; `subject` etc can be anything valid as input to `create_rdf_node`  typically an (NS, local) pair or string literal. Any or all may be `None` (if all are then every triple returned). Etc.

cellmlmanip/model.py
```diff
@@ -555,21 +558,29 @@ def get_symbol_by_ontology_term(self, namespace_uri, local_name):
...
+        subject = create_rdf_node(subject)
+        predicate = create_rdf_node(predicate)
+        object = create_rdf_node(object)
+        return self.rdf.triples((subject, predicate, object))
+
+    def get_rdf_value(self, subject, predicate):
+        """Get the value of an RDF object connected to ``subject`` by ``predicate``."""
```
jonc125: Add to docstring (again, in para after blank line) that exactly one triple must match, and the resulting object should be a literal, the string value of which will be returned.

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,24 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
...
 
-def create_rdf_node(namespace_uri, local_name):
-    """Creates and returns an RDF node ``{namespace_uri}local_name``."""
-    # Ensure namespace prefix can be appended to
-    if namespace_uri[-1] not in ['#', '/']:
-        namespace_uri = namespace_uri + '#'
-
-    return rdflib.Namespace(namespace_uri)[local_name]
+def create_rdf_node(node_content):
+    """Creates and returns an RDF node
```
jonc125: Add full stop :)

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,24 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
...
+
+    :param symbol: node_content EITHER a string or a tuple (namespace_uri, local_name)"""
```
jonc125: Should be `:param node_content: either ...` I think? Perhaps also mention that if a string an RDF literal will be returned. In fact, you could just copy most of the pycml docstring:
```
    Creates an RDF node from the given ``node_content``.

    The ``node_content``, if given, must either be a :class:`rdflib.term.Node`
    instance, a tuple ``(namespace, local_name)``, or a string, in which case
    it is interpreted as a literal RDF node.

    The ``node_content`` may also be `None` to return `None`, allowing easy
    handling of wildcard options to queries.
```

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,24 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
...
+    if node_content is None:
+        return None
+    elif isinstance(node_content, rdflib.Literal) or isinstance(node_content, rdflib.BNode) or \
```
jonc125: Just `isinstance(node_content, rdflib.term.Node):` should work.

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,24 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
...
+            isinstance(node_content, rdflib.URIRef):
+        return node_content
+
+    if not isinstance(node_content, str) and len(node_content) == 2:
```
jonc125: `if isinstance(node_content, tuple):`

cellmlmanip/rdf.py
```diff
@@ -63,10 +63,24 @@ def create_rdf_node(self, node_content=None, fragment_id=None):
...
+        local_name = node_content[1]
```
jonc125: A bit safer (and more compact) is `uri, local_name = node_content`.

tests/test_rdf.py
```diff
@@ -18,24 +19,36 @@ def load_model(name):
...
+    return model
+
+
+@pytest.fixture(scope="module")
+def test_bad_annotations():
+    model = load_model('test_bad_annotations')
+    return model
+
+
 def test_create_rdf_node():
```
jonc125: Should really add some more low-level tests here showcasing the different possible forms for `node_content`. Maybe also check the return types are as expected?

tests/test_rdf.py
```diff
@@ -82,12 +95,43 @@ def test_get_ontology_terms_by_symbol():
...
+        named_attributes.append({'name': name, 'value': value})
+
+    assert len(named_attributes) == 1
+    assert named_attributes[0]['name'] == 'SuggestedForwardEulerTimestep'
+    assert named_attributes[0]['value'] == '0.0002'
+
+    # subject as string
+    named_attrib = []
+    named_attrs = model.get_rdf_annotations(subject='#test_simple_odes', predicate=(PYCMLMETA, 'named-attribute'))
+    for s, p, attr in named_attrs:
```
jonc125: Could just skip this bit and check the list equals `named_attrs` from above?

cellmlmanip/rdf.py
MauriceHendrix: :(

cellmlmanip/rdf.py
MauriceHendrix: Aah that's what I was looking for

cellmlmanip/rdf.py
MauriceHendrix: is that really safer?

cellmlmanip/rdf.py
MauriceHendrix: fine

cellmlmanip/rdf.py
MauriceHendrix: fine

tests/test_rdf.py
MauriceHendrix: Nope they are generators (`<generator object Graph.triples at 0x06282230>`) and are both different. If I turn them into a list the 1st one is empty (as it has already been walked through).
I could turn both into a list before walking through but don't want to give people the idea that that is somehow needed.

tests/test_rdf.py
MauriceHendrix: ok

cellmlmanip/rdf.py
MauriceHendrix: That's not necessarily true string starting with # => URIref

cellmlmanip/model.py
MauriceHendrix: o yes

Issue #93, opened 2019-10-01
Improve docstrings and create API documentation


MichaelClerx:
Todo:

- [x] Add rst docs, add a `[docs]` option in setup.py to install sphinx
- [x] Add some initial files, covering the main classes
- [x] Check all docstrings are capitalised, punctuated, and follow the format (1 sentence w short explanation + blank line + additional info), with parameters ~given in numpy style (seems to be the most common in the docstrings at the moment)~
- [x] Set up automatic doc building and hosting with readthedocs.io ?

skeating:
@MichaelClerx can you paste a 'gold standard'  example so we are all on the same page

MichaelClerx:
For PINTS we've just started using something like this:

```
def eat_biscuits(self, n, tea_factory=None):
    Attempts to eat ``n`` biscuits from the central biscuit repository.
    
    Biscuits will be eaten from best to worst, in the order defined by
    :meth:`cellmlmanip.order_biscuits`, using the algorithm defined in [1]_. An
    optional :class:`cellmlmanip.TeaFactory` can be passed in which may improve
    performance.
        
    Parameters
    ----------
    n : int
        The maximum number of biscuits to eat.
    tea_factory
        An optional :class:`cellmlmanip.TeaFactory` (or ``None``) to aid in
        digesting particularly dry biscuits.
    
    Returns
    -------
    n_eaten : int
        The actual number of biscuits eaten.
    rating : dict
        A dictionary mapping eaten :clas:`cellmlmanip.Biscuit` objects to a
        rating.
    
    References
    ----------
    .. [1] An improved biscuit eating algorithm.
           Clerx (2019) Computational Statistics.
           https://doi.org/10.1007/s001800050022
```

where all parts except the first line are optional:

- Parameter lists only make sense if there are several parameters, or if they have tricky meanings
- Parameter types are optional, but nice
- Return types are usually obvious enough as well, except if you return tuples
- References are nice if you're using some complicated algorithm

MichaelClerx:
Another possible syntax for args and returns is the "google syntax":

```
Args:
    path (str): The path of the file to wrap
    field_storage (FileStorage): The :class:`FileStorage` instance to wrap
    temporary (bool): Whether or not to delete the file when the File
       instance is destructed

Returns:
    BufferedFileStorage: A buffered writable file descriptor
```

[...]

PR #119, opened 2019-11-15
Set up sphinx documentation

MichaelClerx:
First files to allow documentation to be built. Part of #93


MichaelClerx:
We can use readthedocs.io for that, but would make more sense to wait until we have some users I guess

jonc125:
I think you can do it without having to publish on PyPI though, right? And it'll be useful for us to read even if no-one else looks at it!

MichaelClerx:
Yeah you can hook it up to a repo, but I don't really want to add more stuff to my personal account!

jonc125:
I'll do it on mine next week some time then :)

MichaelClerx:
Thanks!

PR #126, opened 2019-11-19
Add latest docs badge

jonc125:


jonc125:
Looks like we'll need to alter the docs config slightly before it'll build anything useful though!

MichaelClerx:
I think we can add a docs.txt requirements file and then tell readthedocs to use it (on the advanced project settings page). Same page lets you set the default branch to `develop`

jonc125:
OK, this now seems to build OK - results are at https://cellmlmanip.readthedocs.io/en/readthedocs/index.html for this branch.

PR #139, opened 2019-11-20
I was using this method saw the renaming but missed it became private

MauriceHendrix:
I did see the renaming but missed that it became private in last review.

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -415,7 +415,7 @@ class Model(object):
             lhs = equation.lhs
 
             # for each of the symbols or derivatives on the rhs of the equation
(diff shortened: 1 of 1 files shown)


MichaelClerx:
Where is that? Can't see it in cg

MauriceHendrix:
> Where is that? Can't see it in cg

`_format_state_variables`

MauriceHendrix:
> Fine by me, feels like if you need this it might be something worth pulling into cellmlmanip, but can't find it just now :-)

what I'm using it for is to figure out which of the state variables I need for the ionic euqations as it might not be all of them, so given a bunch of equations which state vars are in there. Right now I just get all variables for that bunch of equations, which is fine since I'm just looking at state vars anyway.

PR #140, opened 2019-11-21
Find symbols and derivatives

MauriceHendrix:
Also needed to be able to deal with a list

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -415,7 +415,7 @@ class Model(object):
             lhs = equation.lhs
 
             # for each of the symbols or derivatives on the rhs of the equation
-            for rhs in self._find_symbols_and_derivatives(equation.rhs):
+            for rhs in self.find_symbols_and_derivatives(equation.rhs):
                 # if the symbol maps to a node in the graph
                 if rhs in graph.nodes:
                     # add the dependency edge
@@ -694,7 +694,7 @@ class Model(object):
         """
         return float(self.dummy_metadata[symbol].initial_value)
 
-    def _find_symbols_and_derivatives(self, expression):
+    def find_symbols_and_derivatives(self, expression):
         """ Returns a set containing all symbols and derivatives referenced in an expression. """
         symbols = set()
 
@@ -704,7 +704,7 @@ class Model(object):
         # otherwise, descend into sub-expressions and collect symbols
         else:
             for arg in expression.args:
-                symbols |= self._find_symbols_and_derivatives(arg)
+                symbols |= self.find_symbols_and_derivatives(arg)
 
         return symbols
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -1,6 +1,6 @@
 ﻿"""Classes to represent a flattened CellML model and metadata about its variables"""
 import logging
-from collections import OrderedDict, Iterable
+from collections import Iterable, OrderedDict
 from io import StringIO
 
 import networkx as nx
@@ -765,4 +765,3 @@ class Model(object):
 
         # Invalidate cache
         self.graph = None
-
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -689,17 +689,23 @@ def get_value(self, symbol):
...
+
         :param symbol: Sympy Dummy object of required symbol
         :return: float of initial value
         """
         return float(self.dummy_metadata[symbol].initial_value)
 
     def find_symbols_and_derivatives(self, expression):
-        """ Returns a set containing all symbols and derivatives referenced in an expression. """
-        symbols = set()
+        """ Returns a set containing all symbols and derivatives referenced in an expression or list of expressions.
```
jonc125: Normally I'd prefer always to accept just a list, but I see the method is recursive so not that simple. Still, having an argument called `expression` that could be `expressions` feels wrong!

tests/test_model_api.py
```diff
@@ -98,3 +98,11 @@ def test_set_equation(self, model):
         rhs = model.add_number(number=sp.Float(12), units=str(v_unit))
         model.set_equation(lhs, rhs)
 
+    def test_find_symbols_and_derivatives(self, model):
+        model.get_equation_graph()
+        t = model.get_free_variable_symbol()
+        v = model.get_symbol_by_ontology_term(OXMETA, "membrane_voltage")
+        assert model.find_symbols_and_derivatives(t) == model.find_symbols_and_derivatives([t])
+        s = model.find_symbols_and_derivatives(t)
```
jonc125: Could put this definition before the line above to avoid calling `model.find_symbols_and_derivatives(t)` twice?

cellmlmanip/model.py
MauriceHendrix: That's what it used to be before though

MauriceHendrix: Hopefully it's better now

MichaelClerx: It only accepted lists before, but I changed that because I didn't see that functionality used. We can just change it back to what it was before!

cellmlmanip/model.py
```diff
@@ -689,17 +689,23 @@ def get_value(self, symbol):
...
 
+        :param expression: expression or list of expressions to get symbols for
+        """
+        symbols = set()
+        if isinstance(expression, list):
```
MichaelClerx: This only accepts lists, not sequences in general, so usually a bad idea to check like this

cellmlmanip/model.py
MichaelClerx: I'll stick the old code back in, but then once the chaste_cg PR is ready for review again we should take a long time to discuss which bits can go into cellmlmanip. To me this kind of access feels to "involved" for the user code, but I could be very wrong!

cellmlmanip/model.py
MauriceHendrix: Changed it to Iterable

MichaelClerx: hold on hold on, let's just use the code we had before!

cellmlmanip/model.py
MichaelClerx: > I'll stick the old code back in

Now implemented in #143

MauriceHendrix: Sure so what I'm doing thee is the following:
- create a list of dicts with for for each state var (for chaste output) such as name,  units etc.

However the chaste code has 2 different states of equations that are both preceded by **only** the state vars that are used in those equations (not all ll state vars are necessarily used in both).

Alternatives could be something like 
`state_vars_in([exp])` or `contains_var([expr], var)`

jonc125: From what I remember, pycml does the equivalent of `get_equations_for` to get all the equations, and then filters the full state var list accordingly using a set intersection. Not sure if the same approach would be easy with cellmlmanip as it's implemented currently?

MichaelClerx: Moving to #144 !

PR #141, opened 2019-11-21
fixed float precision error on conversion factor and added equivalent_units method

MauriceHendrix:
As I'm currently doing a lot of
if model._get_conversion_factor (..) != 1.0 to avoid generating 1.0* expr

### DIFF
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -381,22 +381,6 @@ class UnitStore(object):
         else:
             return conversion_factor
 
-    def equivalent_units(self, unit_or_expr1, unit_or_expr2):
-        """Returns whether 2 expressions or units are in the same units.
-
-        :param expr_or_unit1: A unit or an expression to look at the units for
-        :param expr_or_unit2: A unit or an expression to look at the units for
-
-        :return: are both parameters in equivalent units (conversion factor == 1.0)
-        """
-        try:
-            print(isinstance(unit_or_expr1, sympy.Expr))
-            unit1 = self.summarise_units(unit_or_expr1) if isinstance(unit_or_expr1, sympy.Expr) else unit_or_expr1
-            unit2 = self.summarise_units(unit_or_expr2) if isinstance(unit_or_expr2, sympy.Expr) else unit_or_expr2
-            return self.get_conversion_factor(unit1, from_unit=unit2) == 1.0
-        except pint.errors.DimensionalityError:
-            return False
-
     def dimensionally_equivalent(self, symbol1, symbol2):
         """Returns whether two expressions, symbol1 and symbol2,
          are dimensionally_equivalent (same units ignoring a calling factor).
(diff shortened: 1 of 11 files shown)


cellmlmanip/units.py
```diff
@@ -368,13 +368,34 @@ def get_conversion_factor(self,
...
+    def equivalent_units(self, unit_or_expr1, unit_or_expr2):
+        """Returns whether 2 expressions or units are in the same units.
+
+        :param expr_or_unit1: A unit or an expression to look at the units for
+        :param expr_or_unit2: A unit or an expression to look at the units for
+
+        :return: are both parameters in equivalent units (conversion factor == 1.0)
+        """
+        try:
+            print(isinstance(unit_or_expr1, sympy.Expr))
```
jonc125: Get rid of this line :)

tests/test_unit_conversion.py
```diff
@@ -1,7 +1,5 @@
 import os
-
```
jonc125: I don't think isort will like this!

tests/test_unit_conversion.py
```diff
@@ -89,3 +87,18 @@ def test_conversion_factor_same_units(simple_model):
     assert simple_model.units.get_conversion_factor(to_unit=to_unit, from_unit=from_unit) == 1
     # expression to unit
     assert simple_model.units.get_conversion_factor(to_unit=to_unit, expression=expression) == 1
+
+
+def test_equivalent_units(model):
+    time_var = model.get_symbol_by_ontology_term(OXMETA, 'time')
+
+    assert model.units.equivalent_units(time_var, model.units.ureg.ms)
+    assert model.units.equivalent_units(time_var, model.units.ureg.ms)
```
jonc125: Why is this line repeated?

cellmlmanip/units.py
```diff
@@ -368,13 +368,34 @@ def get_conversion_factor(self,
...
             assert isinstance(expression, sympy.Expr), 'expression must be of type Sympy expression'
-            return self.convert_to(1 * self.summarise_units(expression), to_unit).magnitude
+            conversion_factor = self.convert_to(1 * self.summarise_units(expression), to_unit).magnitude
+
+        if math.isclose(conversion_factor, 1.0):
+            return 1.0
+        else:
+            return conversion_factor
+
+    def equivalent_units(self, unit_or_expr1, unit_or_expr2):
```
MichaelClerx: Is this the same as `is_unit_equal` ?

tests/test_unit_conversion.py
```diff
@@ -89,3 +87,18 @@ def test_conversion_factor_same_units(simple_model):
...
+    assert model.units.equivalent_units(model.units.ureg.ms, time_var)
+
+    assert not model.units.equivalent_units(model.units.ureg.ms, model.units.ureg.second)
```
jonc125: Aren't the pure units cases tested below? There's definitely some test duplication here!

cellmlmanip/units.py
MauriceHendrix: Actually yes I didn't notice, let me remove my new one.

tests/test_unit_conversion.py
MauriceHendrix: it's not something I changed but we'll see

PR #145, opened 2019-11-21
Fixed bug in Model.get_value() introduced with changes to Model.graph

MichaelClerx:
And added test.


MichaelClerx:
Hey @MauriceHendrix ! Can you have a quick look so cg passes tests again?

The change is that before `get_value()` used the graph with sympy.Float objects, but now it uses the one with our numbers & units

PR #150, opened 2019-11-25
files to test bug

### LINKED ISSUE #121: Possible bug with identifying free variable
I was using a very basic file to test model functions I hit the apparently unreachable code which says I have no free variable in the model (line 520 model.py)

Stepping through the code that looks for the free variable loops through the nodes of the graph. However, when creating the graph, a variable (state/free) of a derivative is only added as a node if it occurs on the rhs of the derivative; which since model uses dsv11/dt = 1 't' does not appear on rhs. So the Model object marks it as a free variable BUT no graph node ever gets created so it cannot be looked up.

Is this a bug ?

skeating:
Fixes #121

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -458,6 +458,12 @@ class Model(object):
                     graph.add_node(rhs, equation=sympy.Eq(rhs, dummy), variable_type='parameter')
                     graph.add_edge(rhs, lhs)
 
+            # check that the free variable if defined as a node
+            if lhs.is_Derivative:
+                free_symbol = lhs.variables[0]
+                if free_symbol not in graph.nodes:
+                    graph.add_node(free_symbol, equation=None, variable_type=free_symbol.type)
+
         # Add more meta-data to the graph
         for variable in graph.nodes:
             if not variable.is_Derivative:
--- a/tests/test_model_functions.py
+++ b/tests/test_model_functions.py
@@ -1,0 +1,49 @@
+import os
+from collections import OrderedDict
+
+import pytest
+
+import cellmlmanip
+import cellmlmanip.rdf
+from cellmlmanip.units import UnitStore
+
+OXMETA = "https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#"
+
+
+class TestModelFunctions():
+    # These represent CellML <units><unit>...</unit></units> elements
+    test_definitions = OrderedDict({
+        'ms': [{'units': 'second', 'prefix': 'milli'}],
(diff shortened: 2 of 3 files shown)


jonc125:
The 'fixes #121' should go in the issue description, not the title.

tests/test_model_functions.py
```diff
@@ -0,0 +1,59 @@
...
+            {'units': 'mV', 'exponent': '1'}, {'prefix': 'micro', 'units': 'second', 'exponent': '-1'}],
+        'mM_per_ms': [{'units': 'mM'}, {'units': 'ms', 'exponent': '-1'}],
+        'ms_power_prefix': [{'prefix': '-3', 'units': 'second'}],
+        'ms_with_multiplier': [{'multiplier': 0.001, 'units': 'second'}],
+    })
+
+    @pytest.fixture
+    def model(scope='class'):
+        return cellmlmanip.load_model(
+            os.path.join(os.path.dirname(__file__), 'cellml_files', "basic_ode.cellml"))
```
MichaelClerx: Can we use `'` for both strings? :D

tests/test_model_functions.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @pytest.fixture
+    def other_model(scope='class'):
+        return cellmlmanip.load_model(
+            os.path.join(os.path.dirname(__file__), 'cellml_files',
+                         "aslanidi_model_2009.cellml"))
```
MichaelClerx: same

tests/test_model_functions.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # also tested in test_hodgkin
+    def test_get_state_symbols(self, model):
+        state_symbols = model.get_state_symbols()
+        assert len(state_symbols) == 1
```
MichaelClerx: Test that it's the correct one too?

cellmlmanip/model.py
```diff
@@ -458,6 +458,12 @@ def graph(self):
                     graph.add_node(rhs, equation=sympy.Eq(rhs, dummy), variable_type='parameter')
                     graph.add_edge(rhs, lhs)
 
+            # check that the free variable if defined as a node
```
MichaelClerx: Could we move this up to around line 428?

skeating: Might get added twice - since it may occur on rhs of another equation and get added there

MichaelClerx: But you could do the same check as here, no?

MichaelClerx: oh ok get it now

Issue #14, opened 2017-11-17
Copyright ModellingWebLab


MichaelClerx:
The license attributes the copyright to "ModellingWebLab", shouldn't we specify this a bit more accurately?

jonc125:
Yeah, now we have a CLA signed between Oxford & UCL we should just use the same approach as Chaste.

jonc125:
See also https://github.com/ModellingWebLab/WebLab/blob/master/LICENSE (which needs updating to 2018...)

MichaelClerx:
> we should just use the same approach as Chaste.

Does that mean writing copyright oxford as in the web lab example?

PR #151, opened 2019-11-25
Updated license

MichaelClerx:
See #14


jonc125:
Could we get into the habit of using `fixes` etc in issue descriptions?

MichaelClerx:
Sure! I thought it only picked those up in commit messages though

jonc125:
GitHub picks them up from both places. I didn't notice it in the commit comment.

PR #155, opened 2019-11-25
Moved transpiler out of its own module

### LINKED ISSUE #147: Move transpiler module outside of its own directory + simplify it
- [x] move into cellmlmanip root module
- [x] remove unused options

MichaelClerx:
Fixes #147


MichaelClerx:
Most of the simplification already happened in #148.
There's still some code that's only used in testing:
- The option to let the transpiler create its own symbols & number objects
- The option to have mathml numbers without cellml units

Not sure if we should remove this?

jonc125:
I'd merge this but not do anything further on the transpiler for now.

PR #157, opened 2019-11-26
Need to be able to tell if equation units are equivalent to a specific unit, so that I can find ionic currents

MauriceHendrix:


MichaelClerx:
This doesn't make sense to me. Why use the same method for 2 different things??

MauriceHendrix:
> This doesn't make sense to me. Why use the same method for 2 different things??

it's not 2 different things at all

MichaelClerx:
The original method has badly named variables too. Why are we calling an expression a "symbol". A symbol is a type of expression, not the other way around.

And what's wrong with doing summarise_units() yourself and then comparing the resulting unit objects? Does pint not have a method for that?

MauriceHendrix:
> The original method has badly named variables too. Why are we calling an expression a "symbol". A symbol is a type of expression, not the other way around.
> 
> And what's wrong with doing summarise_units() yourself and then comparing the resulting unit objects? Does pint not have a method for that?

what has print got to do with it? I'm looking for currents.
Before I was looking for things dimensionally_equivalent to the _membrane_stimulus_current_units.
But now I'm meeting models without _membrane_stimulus_current_units, and there are set units to try in that case, but the method doesn't like just units. 

That's what this is fixing.

jonc125:
pint didn't have a method to check directly if units are in the same dimension. The typical approach is to try calling `.to` and catching `DimensionalityError`, which is what this method does. I agree about the poor argument naming though!

Although looking at https://pint.readthedocs.io/en/latest/developers_reference.html things may have improved, e.g. `Unit.dimensionality` or `Quantity.check`.

cellmlmanip/units.py
```diff
@@ -388,16 +388,19 @@ def get_conversion_factor(self,
         else:
             return conversion_factor
 
-    def dimensionally_equivalent(self, symbol1, symbol2):
-        """Returns whether two expressions, symbol1 and symbol2,
+    def dimensionally_equivalent(self, units_or_symbol1, units_or_symbol2):
+        """Returns whether two units or expressions, units_or_symbol1 and units_or_symbol2,
```
MichaelClerx: Can we change the parameters to just be called `x` and `y`?

cellmlmanip/units.py
```diff
@@ -388,16 +388,19 @@ def get_conversion_factor(self,
...
          are dimensionally_equivalent (same units ignoring a calling factor).
```
MichaelClerx: calling --> scaling ?

cellmlmanip/units.py
```diff
@@ -388,16 +388,19 @@ def get_conversion_factor(self,
...
-        :param symbol1: the first expression to compare
-        :param symbol2: the second expression to compare
+        :param units_or_symbol1: the first units or expression to compare
+        :param units_or_symbol2: the second units or expression to compare
         :return True if units are equal (regardless of quantity), False otherwise
         """
         try:
-            self.get_conversion_factor(from_unit=self.summarise_units(symbol1),
-                                       to_unit=self.summarise_units(symbol2))
+            unit1 = units_or_symbol1 if isinstance(units_or_symbol1, self.ureg.Unit) \
```
MichaelClerx: Could we do a
```
if isinstance(sympy.Basic):
    x = self.summarise_units(x)
```
instead?

cellmlmanip/units.py
```diff
@@ -388,16 +388,19 @@ def get_conversion_factor(self,
...
+                else self.summarise_units(units_or_symbol1)
+            unit2 = units_or_symbol2 if isinstance(units_or_symbol2, self.ureg.Unit) \
+                else self.summarise_units(units_or_symbol2)
+            self.get_conversion_factor(from_unit=unit1, to_unit=unit2)
```
MichaelClerx: Should this line be the only line in the try-catch? (With the rest coming before it)

MauriceHendrix:
> mV_per_ms.dimensionality == A_per_F.dimensionality

actually yes that works!
it also works for:
'nano_mole_per_cubic_mm': [{'units': 'mole', 'prefix': 'nano"}, {'units': 'metre', 'prefix': 'milli',, 'exponent': '-3'}]
'millimolar': [{'units': 'mole', 'prefix': 'milli'}, {'units': 'litre', 'exponent': '-1'}]

so it looks like we don't need this method anymore at all.

jonc125:
Looks like pint does work as we'd want:
```python
>>> from pint import UnitRegistry
>>> ureg = UnitRegistry()
>>> A_per_F = ureg.ampere / ureg.farad
>>> V_per_s = ureg.volt / ureg.second
>>> A_per_F, V_per_s
(<Unit('ampere / farad')>, <Unit('volt / second')>)
>>> V_per_s.dimensionality == A_per_F.dimensionality
True
>>> V_per_s.dimensionality
<UnitsContainer({'[current]': -1.0, '[length]': 2.0, '[mass]': 1.0, '[time]': -4.0})>
```

Issue #135, opened 2019-11-20
Setting the initial value of a variable that was just upgraded to a state


MichaelClerx:
Say I upgrade a variable to a state variable (by replacing its defining equation).
How do I then go about setting its initial value?
Are we expecting users to manipulate the dummy info objects themselves?

jonc125:
I'm not sure requiring users to do that would be wise! Although depends on how clear that API is. In pycml you can just do `variable.initial_value = "..."`; something equally simple in cellmlmanip would be nice.

MichaelClerx:
It would! But we're stuck with symbols + meta-info-in-the-model because of the no subclassing thing. So it'd have to be e.g. `Model.set_initial_value(symbol, value)`

jonc125:
I'd be OK with that API though :)

jonc125:
Can we now close this?

MichaelClerx:
I'm not sure! There's still no mechanism that ensures only states have initial values. Is that something we want?

jonc125:
What about constants? I can't recall if they're kept as variables with initial values, or whether they are given assignment equations instead?

MichaelClerx:
@tamuri ?

tamuri:
In the method that builds the equation graph, we iterate over all the symbols on the rhs. If it's not a state or free variable (nor a number dummy), I called it "parameter" (not sure if that's right though) and add an equation assigning that symbol to its initial value - but only on the graph. It's not added as an equation to the model.

MichaelClerx:
Hmmm! For FC and CG I would want them all to act exactly the same. So every variable has 1 defining equation that I can mess around with, and it shouldn't matter if it's a state, parameter, etc.
The whole 'defining constants through an initial value' in CellML seems like a mistake to me

MichaelClerx:
So for the original question... We do expect users now to say `dummy.initial_value = 12` ?

MichaelClerx:
Similarly, after turning a state into a non-state we have to remove the initial value?

(Although post-parsing this sort of thing isn't checked)

jonc125:
Yep.

PR #158, opened 2019-11-26
get_symbol_by_cmeta_id bug

### LINKED ISSUE #156: Possible bug in get_symbol_by_cmeta_id
Again using my very basic model:

    <component name="env_ode">
        
        <variable name="time" public_interface="in" units="ms"/>
        <variable cmeta:id="sv11" initial_value="2" name="sv1" units="mV"/>
        <math xmlns="http://www.w3.org/1998/Math/MathML">
            <apply>
                <eq/>
                <apply>
                    <diff/>
                    <bvar>
                        <ci>time</ci>
                    </bvar>
                    <ci>sv1</ci>
                </apply>
                <cn cellml:units="mV_per_ms">1</cn>
            </apply>
        </math>
    </component>

no node is every created for the variable sv1 and so get_symbol_by_cmeta_id('sv11')  does not find the variable.

The model is so simple it does not ever occur as lhs of anything but it would be the variable that was the output of simulating this toy model.

Bug or not ??

skeating:
Fixes #156

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -456,11 +456,14 @@ class Model(object):
                     graph.add_node(rhs, equation=sympy.Eq(rhs, dummy), variable_type='parameter')
                     graph.add_edge(rhs, lhs)
 
-            # check that the free variable if defined as a node
+            # check that the free  and state variables are defined as a node
             if lhs.is_Derivative:
+                state_symbol = lhs.free_symbols.pop()
                 free_symbol = lhs.variables[0]
                 if free_symbol not in graph.nodes:
                     graph.add_node(free_symbol, equation=None, variable_type=free_symbol.type)
+                if state_symbol not in graph.nodes:
+                    graph.add_node(state_symbol, equation=None, variable_type=state_symbol.type)
 
         # Add more meta-data to the graph
         for variable in graph.nodes:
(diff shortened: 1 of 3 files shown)


MichaelClerx:
@MauriceHendrix what do you think? I guess you never use the graph directly, so the state variables being in it won't bother you either?

MichaelClerx:
We could also consider sticking the initial value in as the equation, I guess?

jonc125:
> We could also consider sticking the initial value in as the equation, I guess?

Don't do this, it'll lead to an over-constrained model!

MauriceHendrix:
> > We could also consider sticking the initial value in as the equation, I guess?
> 
> Don't do this, it'll lead to an over-constrained model!
> 
> While as suggested elsewhere adding equations for constants makes sense, we do need to keep the initial_value concept for state variables.

I would then need to filter out these initial values as they're treated differently in the output, not impossible, but the way it is it's quite convenient.

jonc125:
> > > We could also consider sticking the initial value in as the equation, I guess?
> > 
> > 
> > Don't do this, it'll lead to an over-constrained model!
> > While as suggested elsewhere adding equations for constants makes sense, we do need to keep the initial_value concept for state variables.
> 
> I would then need to filter out these initial values as they're treated differently in the output, not impossible, but the way it is it's quite convenient.

Which initial values do you mean? See also #135, particularly the last comment.

PR #162, opened 2019-11-30
added keep_unused_eqs flag to model.get_eqiations_for

MauriceHendrix:


jonc125:
I don't understand why this is needed! Surely the point of `get_equations_for` is that it only returns the equations needed to calculate the symbols provided, so there won't be any unused ones? If you want other equations, you can look at the full list or graph.

MauriceHendrix:
> I don't understand why this is needed! Surely the point of `get_equations_for` is that it only returns the equations needed to calculate the symbols provided, so there won't be any unused ones? If you want other equations, you can look at the full list or graph.

Well this isn't the case. Some example output from the `espinosa_model_1998_normal` model from cahste_cg and pycml. Both cause a compile error (unused variable):
**chaste_cg:**
```
const double var_calcium_release__VoltDep = 0.040762203978366204 * exp(0.08 * var_chaste_interface__membrane__V);
const double var_calcium_release__ActRate = 500.0 * pow(var_calcium_release__RegBindSite, 2.0);
```
**pycml generates:**
```
const double var_calcium_release__V = var_chaste_interface__membrane__V; // millivolt
....
const double var_calcium_release__VoltDep = exp(0.080000000000000002 * (var_calcium_release__V - 40.0));
....
const double var_calcium_release__ActRate = (0.0 * var_calcium_release__VoltDep) + (500.0 * pow(var_calcium_release__RegBindSite, 2.0));
```

jonc125:
Hmm, that sounds like a more serious underlying issue, not something to be fixed with a hack. `cellmlmanip` shouldn't include the VoltDep equation if it isn't used - that's the whole point of how the graph is created! If VoltDep is there, *something* in the generated equations referenced it. Are you simplifying the equations after substituting dummies or something, perhaps? Given the difference in how it's defined, looks like it!

MauriceHendrix:
Didn't realise the current behaviour wasn't the desired one. I'm not explicitly simplifying anything. As you can see from my fix /hack it just goes through the resulting eqs and removes unused ones without simplifying anything. I suspect when generating the graph the generation is based on the terms read from cellml and maybe it should read them from the created symptoms eq instead?

MauriceHendrix:
You can probably fix it in the parser but it wouldn't be in the graph. 

You could do it during the getting of ancestors and check there not sure that's necessarily more efficient than doing it once at the end as you don't have the equations. Equally if doing it when wedding eqs, you need to go through eqs so far to check for every eq.

jonc125:
Hmm. I'm just responding by email so can't see the code at the mo. But what the graph generation *should* be doing is creating an edge between terms if the definition of the source uses the destination on the RHS. So by *definition* following edges from a list of equations will *only* give you those equations needed to calculate the inputs. If it's not doing that, something is going screwy somewhere!

MauriceHendrix:
Yeah it's of course just symptoms simplifying things early. If that's what we definitely want, we should change parsing to check free symbols in eqs before adding ancestors to the graph. 

Just checking for 0* wouldn't be enough as there are other ways to get the same. Bit of a silly model defenition but o well.

cellmlmanip/model.py
```diff
@@ -207,6 +208,8 @@ def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True, st
         :param recurse: indicates whether to recurse the equation graph, or to return only the top level equations
         :param strip_units: if ``True``, all ``sympy.Dummy`` objects representing number with units will be replaced
             with ordinary sympy number objects.
+        :param: keep_unused_eqs indicates whether any unsued equations  should be kept. If keep_unused_eqs==False
```
skeating: unsued => unused

cellmlmanip/model.py
```diff
@@ -207,6 +208,8 @@ def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True, st
...
+            only equations for which the lhs is used in the rhs of any equation, or is in the `symbols` input are kept.
```
skeating: I'm not clear what the distinction between keeping unused equations and not keeping them actually is

cellmlmanip/model.py
MauriceHendrix: Problem is sympy simplifying out terms e.g. 0*x +y -> y which leads to the equation (for x) being there but the lhs not being used anywhere, see above for popper example.

cellmlmanip/model.py
MauriceHendrix: Problem is sympy simplifying out terms e.g. 0*x +y -> y which leads to the equation (for x) being there but the lhs not being used anywhere, see above for popper example.

jonc125:
Sorry this is getting frustrating @MauriceHendrix ! Clearly this has picked up a cellmlmanip bug that we should have spotted in earlier testing. As such, I'd rather not paper it over with a quick fix, but will try to have a proper look to understand what's going on later today. In the meantime, if this is a blocker for you, I'd suggest that Chaste cg filters the list it gets from `get_equations_for` as a temporary measure. Does that seem reasonable?

MauriceHendrix:
Ok!

tamuri:
It sounds like this happens because when the graph is first built, the zero is a dummy and VoltDep term remains in the rhs in that equation. Once the zero is substituted for a number, Sympy removes VoltDep from the equation but the edge is still there...?

jonc125:
> It sounds like this happens because when the graph is first built, the zero is a dummy and VoltDep term remains in the rhs in that equation. Once the zero is substituted for a number, Sympy removes VoltDep from the equation but the edge is still there...?

Yeah, that looks plausible. `graph_with_sympy_numbers` doesn't change any of the edges in the graph, only substitutes within equations. So a variable might disappear from an equation without removing the corresponding edge. Created #165 to pick up on this.

Can we therefore close this PR @MauriceHendrix ?

PR #167, opened 2019-12-02
Add better tests for get_equations_for, fix bug with simplification

### LINKED ISSUE #165: More robust checking of `get_equations_for`
This arises from #162. It seems that the simplification that happens when `strip_units=True` can result in invalid edges remaining. We need to add more robust tests, and then fix issues revealed; hopefully just by removing edges from `graph_with_sympy_numbers` when it's created!

Ideally we'd have a simple model containing easy to verify test cases:
- Equations something like:
    1. `dy1/dt = number`
    2. `dy2/dt = var1`
    3. `dy3/dt = var2 * (number + dy1/dt)`
    4. `var1 = (5 - 5) * var3`
    5. `var2 = 23 + var4`
    6. `var4 = -23`
    7. `var3 = number / number`
    8. `var5 = var4 + var3`
- Test cases:
    - [x] `g_e_f([dy2/dt], strip_units=False)` -> `[eq7, eq4, eq2]`
    - [x] `g_e_f([dy2/dt], strip_units=True)` -> `[eq4, eq2]` (simplification sets eq4 as `var1=0`)
    - [x] `g_e_f([dy3/dt], strip_units=True)` -> `[eq1, eq6, eq5, eq3]` (because the simplification is only within a single equation, not across the system)
    - [x] `g_e_f([var5], lexicographical_sort=False)` -> `[eq6, eq7, eq8]` (assuming document order if not sorting by lhs name? Not sure if we can guarantee any order though!)
    - [x] `g_e_f([var5], lexicographical_sort=True)` -> `[eq7, eq6, eq8]`
    - [x] Extra cases to test `recurse=False`!

### LINKED ISSUE #166: Remove tests marked as redundant by Sarah
@skeating I think you're right those two tests in test_model_units can go. Mind if I pick this up?

### LINKED ISSUE #168: Do we need the `lexicographical_sort=False` option in `Model.get_equations_for()` ?
Would simplify things if we took it out.

MichaelClerx:
Fixes #165 
Fixes #166
Fixes #168

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -198,13 +198,12 @@ class Model(object):
             rhs_units, self.units.ureg.get_base_units(rhs_units)
         )
 
-    def get_equations_for(self, symbols, lexicographical_sort=True, recurse=True, strip_units=True):
+    def get_equations_for(self, symbols, recurse=True, strip_units=True):
         """Get all equations for a given collection of symbols.
 
-        Results are sorted in topographical order.
+        Results are sorted first by dependencies, then by variable name.
 
         :param symbols: the symbols to get the equations for.
-        :param lexicographical_sort: indicates whether the result is sorted in lexicographical order first.
         :param recurse: indicates whether to recurse the equation graph, or to return only the top level equations.
         :param strip_units: if ``True``, all ``sympy.Dummy`` objects representing number with units will be replaced
             with ordinary sympy number objects.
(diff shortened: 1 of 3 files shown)


MichaelClerx:
First tests recreate the issue

MichaelClerx:
Have added a fix

(It does involve a lot of traversing of expression trees so won't be the fastest method I think)

cellmlmanip/model.py
```diff
@@ -506,6 +506,18 @@ def graph_with_sympy_numbers(self):
 
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
+                # Get RHS with and without substitution
+                rhs1 = equation.rhs
+                rhs2 = rhs1.subs(subs_dict)
+
+                # Check if simplification removed dependencies on other variables, and if so remove the corresponding
+                # edges.
+                refs1 = self.find_symbols_and_derivatives([rhs1])
```
jonc125: Can you get this just from the existing edges instead?

MichaelClerx:
Ready for review now. The test is longer than I would have liked but can't see a way to split it up without significant code duplication

MauriceHendrix:
I don't think you can make the test much shorter other than reading in from a small model file. You could also add a **0  case (even though I'm pretty sure the implementation will catch that).

PR #171, opened 2019-12-02
Replaced Model.set_equation(lhs, rhs) with Model.remove_equation(equation)

### LINKED ISSUE #118: Replace a variable's equation
Given a variable, we need a method to set/replace its RHS

From FC, I can read equations (with variables specified as ontology terms) and units (using an FC unit syntax).

In a two-way communication approach, I could then:

- ask the cellmlmanip model for the symbols for each term
- ask the cellmlmanip model for the unit objects corresponding to my units
- pass in an equation with symbols pointing to model variables and literals as dummies containing the model's unit vars

Buuuut, that would mean fc talking to cellmlmanip through cg, so would probably prefer a one-way approach

- give cellmlmanip a sympy equation with variables as ontology terms and units in some syntax that `pint` understands?
- cellmlmanip then replaces the ontology terms and units

So then we'd need to write some mini-spec for how cellmlmanip expects variable symbols and units?

MichaelClerx:
Fixes #118.

As discussed in #118 I removed `set_equation(lhs, rhs)` and instead added `remove_equation(eq)` which can be used together with `add_equation(eq)` to achieve the same effect.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -112,9 +112,11 @@ class Model(object):
 
         :return: A :class:`VariableDummy` object.
         """
+        # Check for clashes
         if name in self._name_to_symbol:
             raise ValueError('Variable %s already exists.' % name)
 
+        # Add variable
         self._name_to_symbol[name] = var = VariableDummy(
             name=name,
             units=self.units.get_quantity(units),
@@ -124,6 +126,9 @@ class Model(object):
             order_added=len(self._name_to_symbol),
             cmeta_id=cmeta_id,
         )
+
+        # Invalidate cached graphs
+        self._invalidate_cache()
 
         return var
 
@@ -173,12 +178,19 @@ class Model(object):
 
             logger.debug('Updated target: %s', target)
 
+            # Invalidate cached graphs
+            self._invalidate_cache()
+
             return True
 
         # The source variable has not been assigned a symbol, so we can't make this connection
         logger.info('The source variable has not been assigned to a symbol '
                     '(i.e. expecting a connection): %s ⟶ %s',
                     target.name, source.name)
+
+        # Invalidate cached graphs
+        self._invalidate_cache()
+
         return False
 
     def add_rdf(self, rdf: str):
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -569,36 +581,17 @@ def find_symbols_and_derivatives(self, expression):
...
-                i_existing = i
-                break
-
-        # Add or replace equation
-        if i_existing is None:
-            self.equations.append(sympy.Eq(lhs, rhs))
-        else:
-            self.equations[i_existing] = sympy.Eq(lhs, rhs)
+        try:
+            i = self.equations.index(equation)
```
MauriceHendrix: Personally I wouldn't call it i if it's not a loop control variable.

cellmlmanip/model.py
```diff
@@ -569,36 +581,17 @@ def find_symbols_and_derivatives(self, expression):
...
+        except ValueError:
```
MauriceHendrix: Would there ever be a scenario where you may try to remove eqs multiple times? E.g. something like
eqs = model.get_equations_for(model.derivative_symbols)) ? Where it could be possible for a variable to be used in more than 1? 
I know this particular example is little silly but can we think of a legit use case?

cellmlmanip/model.py
```diff
@@ -173,12 +178,19 @@ def connect_variables(self, source_name: str, target_name: str):
...
+
             return True
 
         # The source variable has not been assigned a symbol, so we can't make this connection
         logger.info('The source variable has not been assigned to a symbol '
                     '(i.e. expecting a connection): %s ⟶ %s',
                     target.name, source.name)
+
+        # Invalidate cached graphs
+        self._invalidate_cache()
```
MauriceHendrix: What you could do is pull out the return statement and invalidate cache as that's double and then return  `source.assigned_to and source.units != target.units. I know it doesn't matter and of course return True/False is easier to write but you'd be less likely to accidentally forget to invalidate cache if things change I think

cellmlmanip/model.py
```diff
@@ -569,36 +581,17 @@ def find_symbols_and_derivatives(self, expression):
...
-            if symbol == lhs_symbol:
-                i_existing = i
-                break
-
-        # Add or replace equation
-        if i_existing is None:
-            self.equations.append(sympy.Eq(lhs, rhs))
-        else:
-            self.equations[i_existing] = sympy.Eq(lhs, rhs)
+        try:
```
jonc125: Can't you just do
```python
try:
    self.equations.remove(equation)
except ValueError:
    raise ...
```
?

cellmlmanip/model.py
jonc125: Not sure what you mean here? I think all modification methods should invalidate the cache, as it's too easy to forget otherwise. Unless you mean this method could do so just once, rather than having that line twice?

MauriceHendrix: That's what I mean yeah.

cellmlmanip/model.py
MichaelClerx: I don't think so! The equations all have to have the form `x = ...`. So there can't be two equations for `x`. Variables can be used lots of times in the RHS, but only in one place in the LHS, so there's never a duplicate equation.

cellmlmanip/model.py
MichaelClerx: OK refactoring this method to have nicer flow

cellmlmanip/model.py
MichaelClerx: woops

Issue #132, opened 2019-11-20
Change `units` arguments in Model methods to accept unit objects as well as strings


MichaelClerx:
To get a number in the units of a specific variable, I now need to do:

        model.get_equation_graph()
        v = model.get_symbol_by_ontology_term(OXMETA, 'membrane_voltage')
        v_meta = model.get_meta_dummy(v)
        unit = v_meta.units
        number = model.add_number(sp.Number(-80), str(unit))

might make more sense if I could use the `unit` directly? Or would that add some unforseen complication?

MichaelClerx:
Woops, this was recently discussed in #77

@jonc125 said:
>  The possibly complex bit is as you say if you want a 'nice' name for the units, which only matters either when referring to them by name (which I don't think we should do after the parsing - much easier to refer to actual objects) or when reporting to users

and

> I agree with the suggestion for using a pint units definition when asking for conversion factors, rather than having to register one with the UnitsStore. That seems like quite a long-winded way to get the units for a variable too; passing the units field from the associated MetaDummy might be better? And perhaps a helper method Model.get_units_for(symbol) which does this?

MichaelClerx:
So 

- [ ] Update all methods accepting units to accept unit objects
- [ ] Update them to _only_ accept unit objects, change the parsing code to ask for the unit object that matches a string itself

MichaelClerx:
@MauriceHendrix would this cause trouble for you? We can also think about accepting both strings and unit objects

MauriceHendrix:
as per previous whatever you do with the graph should be ok. I think should probably make it private and use only the api, as otherwsie what's the point of having an api!

MichaelClerx:
Anyone object to having the methods accept both units and strings?

Then this would become

- [ ] Update docstrings to say units & strings are both fine

skeating:
I think that's more usable.

PR #172, opened 2019-12-02
Model.add_number and add_variable now allow unit objects as well as strings

MichaelClerx:
See #132 

Leaving in the option to use strings for now, as it's used in lots of tests etc. Might be nice for users too


cellmlmanip/model.py
```diff
@@ -92,19 +92,24 @@ def add_number(self, value, units):
         Creates and returns a :class:`NumberDummy` to represent a number with units in sympy expressions.
 
         :param number: A number (anything convertible to float).
-        :param units: A string unit representation.
+        :param units: A `pint` units representation
```
jonc125: As this also accepts strings, should we keep that option in the docs too? Or hide it just for use by our tests?

jonc125: Oh, and should it be double backticks? My reST is rusty!

cellmlmanip/model.py
```diff
@@ -387,6 +397,12 @@ def get_ontology_terms_by_symbol(self, symbol, namespace_uri=None):
                         ontology_terms.append(uri_parts[-1])
         return ontology_terms
 
+    def get_units(self, name):
```
jonc125: Should we use this method above?

cellmlmanip/units.py
```diff
@@ -250,10 +250,10 @@ def _define_pint_unit(self, units_name, definition_string_or_instance):
         self.custom_defined.add(units_name)
 
     def get_quantity(self, unit_name):
-        """Returns a pint.Unit with the given name from the UnitRegistry.
+        """Returns a pint `Unit` with the given name from the UnitRegistry.
         :param unit_name: string name of the unit
-        :return: pint.Unit
-        throws pint.UndefinedUnitError if te unit is not present in registry
+        :return: `Unit`
+        throws pint.UndefinedUnitError if the unit is not present in registry
```
jonc125: Could use the Sphinx `:raises pint.UndefinedUnitError: if...` syntax here, actually?

PR #175, opened 2019-12-05
adding unit conversion functions

### LINKED ISSUE #77: Adding unit conversion for  code generation
Code generation needs to be able to convert units. Currently Maurice is doing

factor = model.units.get_conversion_factor(1 * current_units, desired_units)
eq = sp.relational.Equality(eq.lhs, factor * eq.rhs)

which works so long as lhs isnt used elsewhere.

This issue will overlap with parts of other issues e.g. #8, #21 but adding it here to be clear what I'm working on.

skeating:
Fixes #77

## Description
Adding functions necessary for unit conversion.
- [x] take a variable and unit and convert all relevant values/equations in model to reflect new unit on variable
- [x] add_input function
- [x] add_output function
- [x] model missing units - MOVED to issue #192
- [x] arguments as strings/objects
- [x] review arg checking
- [x] refactor is_output to not be in API
- [x] test state symbols are correct after an input is added
- [x] test substitution of rhs deriv recurses into sub expressions
- [x] add helper function for getting unique name
- [x] check _convert_variable_instance need not replace deriv for output (step 4)
- [x] formatting of docstring examples
- [x] change function name/args/add enum

## Motivation and Context
Fixes #77 

## Types of changes

## Checklist:

## Testing


MichaelClerx:
This is looking good!

skeating:
Requesting review of whats here.
Questions: 
1. how do I distinguish between input/output if I wanted to create a generic convert(some_variable, some_units) and it figured out which to call ?
2. should I make it so that arguments are either string/DummyVariable and string/Unit object ?
3.  any ideas why appveyor is failing - tests pass locally
4. I know I need to deal with case where the model does not have the required units in its registry

MichaelClerx:
It's using sympy 1.5, that could be one thing :-)

jonc125:
I'll try to review tomorrow afternoon once back from physio. Quick responses to questions:

1. I don't think we want to do this, as you can't tell inside the function which is wanted, only the caller knows.
2. Probably yes. Certainly we want to be able to pass objects. Being able to do strings as well makes things more consistent.
3. Haven't looked.

MichaelClerx:
> It's using sympy 1.5, that could be one thing :-)

Another bit of evidence for this is that travis passes with pinned requirements, but the one without pinned requirements fails

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
         """ Returns an iterator over this model's variable symbols. """
         return self._name_to_symbol.values()
 
+    def add_input(self, name, units, is_output=False):
```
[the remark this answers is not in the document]
jonc125: Um, I didn't claim that! Or at least I didn't mean to :)

Personally I like the `add_input` and `add_output` names, as they match what I used before and how I think about this. I'm open to others (@MauriceHendrix ?) suggesting something better, but haven't seen a concrete suggestion yet.

From an FC point of view, specifying the variable either by passing a `VariableDummy` or ontology term would be most useful, so those would be the two options I'd suggest for a first argument called `variable`. Giving the units as an object as done here is also what we'd want.

It's possible that Chaste code-gen would need to convert a variable that doesn't have an annotation, so the option of both might help @MauriceHendrix . And even if all variables do have annotations, the caller might well already know the `VariableDummy` so it makes sense to avoid another lookup.

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+        if not original_variable:
+            raise KeyError('No variable with name "%s" found.' % name)
+
+        original_units = original_variable.units
+
+        # no conversion necessary
+        if original_units == units:
+            return
+
+        state_symbols = self.get_state_symbols()
```
jonc125: These 2 lines could go after the `is_output` check below.

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+        free_symbol = self.get_free_variable_symbol()
+        # conversion_factor for old units to new units
+        cf = self.units.get_conversion_factor(from_unit=original_units, to_unit=units)
+
+        # create new variable and relevant equations
+        new_variable = self._convert_variable_instance(original_variable, cf, units)
+
+        # if is output do not need to do additional changes for state/free symbols
+        if is_output:
```
jonc125: Actually, I'm not at all convinced by the idea of having `is_output` as an option to `add_input`! It may be convenient for implementation, but it looks odd as an API :)

Could we instead have common internal methods that both `add_input` and `add_output` call, that give us `units`, `original_variable`, etc? (We already have one-liners to set `cf` and `new_variable`.) Then `add_output` can still be very short and the API is clearer. Things like getting a `Unit` object from either name or `Unit`, getting a `VariableDummy` from object or annotation, etc are probably generally useful in a few places too.

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+        """
+        self.add_input(name, units, is_output=True)
+
+    def _replace_derivatives(self, new_derivative):
+        """
+        Function to replace an instance of a derivative that occurs on the RHS of any equation
+        :param new_derivative: new variable representing the derivative
+        """
+        for equation in self.equations:
+            for argument in equation.rhs.args:
```
jonc125: Will this recurse into sub-expressions? I can never remember!

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+    def _convert_variable_instance(self, original_variable, cf, units):
+        """
+        Internal function to create new variable and an equation for it.
+        :param original_variable: VariableDummy object to be converted
+        :param cf: conversion factor
+        :param units: Unit object for new units
+        :return: the new variable created
+        """
+        # 1. get unique name for new variable
+        new_name = original_variable.name + '_converted'
```
jonc125: This 3-line bit of logic is used at least twice; worth extracting a helper method?

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+        # 4. check whether we had an eqn for original
+        # if so, remove it and replace with new_var = rhs * cf
+        for equation in self.equations:
+            if equation.is_Equality and equation.args[0] == original_variable:
+                expression = sympy.Eq(new_variable, equation.args[1] * cf)
+                self.add_equation(expression)
+                self.remove_equation(equation)
+                break
+
+        # 5. add an equation for original variable
```
jonc125: Ah, so this is always using the 'input' flow, and so isn't suitable for 'output' cases, no?

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+        if is_output:
+            return
+
+        new_derivatives = []
+        # if state variable
+        if original_variable in state_symbols:
+            new_derivatives.append(self._convert_state_variable_deriv(original_variable, new_variable, cf))
+
+        # if free variable
+        if original_variable == free_symbol:
```
jonc125: This is good!

cellmlmanip/model.py
```diff
@@ -619,6 +619,205 @@ def variables(self):
...
+            # for each derivative wrt to free variable add necessary variables/equations
+            current_equations = self.equations.copy()
+            for equation in current_equations:
+                if equation.args[0].is_Derivative:
+                    if equation.args[0].args[1].args[0] == original_variable:
+                        new_derivatives.append(self._convert_free_variable_deriv(equation, new_variable, cf))
+
+        # replace any instances of derivative of rhs of other eqns with new derivative variable
+        for new_derivative in new_derivatives:
+            self._replace_derivatives(new_derivative)
```
jonc125: Incidentally, will `self.get_state_symbols` give the right list after we've done this? Should make sure to test that.

cellmlmanip/model.py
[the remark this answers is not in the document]
skeating: what about 'add_and_convert_input/output'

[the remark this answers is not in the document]
jonc125: Hmm, that gives me some ideas. We could do `convert_variable` (or even `maybe_convert_variable`!), and either have two methods (`convert_variable_as_input`) or have the direction as an argument (`maybe_convert_variable(var, units, direction=DataFlowDirection.TO_MODEL)`) with a `DataFlowDirection` enum having options `TO_MODEL` and `FROM_MODEL`. Does that make the API clearer?

It might also be handy to make the return value either the original or newly added variable, depending on whether a conversion was needed?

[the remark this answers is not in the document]
skeating: I like the idea of some way of telling the user 'nothing changed'

skeating:
Thanks guys
I have added to the to do list in the description and marked issue as resolved if I did this. 
So we do need to agree on naming ...

cellmlmanip/model.py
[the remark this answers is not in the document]
jonc125: @MauriceHendrix @MichaelClerx any preferences on naming?

I propose `maybe_convert_variable(var, units, direction)` with 3 required positional arguments.

cellmlmanip/model.py
skeating: actually they cannot since the model has already done some changes so they fail if left til after 
        new_variable = self._convert_variable_instance(original_variable, cf, units)

cellmlmanip/model.py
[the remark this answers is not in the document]
MichaelClerx: I'm happy to let @skeating and @jonc125 decide the names!

cellmlmanip/model.py
skeating: So if we have state_symbol sv1 in mV and we do add_input(sv1, V) we will get a new variable named 'sv1_converted' which is now the state symbol. Just checking I will still have same number of state symbols, but original symbol_sv1 will no longer be a state var but the new symbol_sv1_converted will be

skeating: This is what current code does - is it right ?

jonc125: Yes, that's correct - if the flow direction is input. If output, then the original state var stays a state var.

MauriceHendrix:
I like the convert_variable(var, units, direction) but probably without the maybe :).

As input I would always already have the Variable dummy so no need for ontology term on my account but I can see how it could be useful to others.

Are we getting anything return? Or will it throw an error when it can't convert?

As for return value I definitely like those ideas. What would you do if it can't be converted (e.g. I ask to convert time in ms to time in kg) throw a dimensionality error?

jonc125:
> I like the convert_variable(var, units, direction) but probably without the maybe :).

It's certainly shorter without the maybe! I'm happy either way, provided the docs make clear that it returns the original variable if no conversion needed.

> As for return value I definitely like those ideas. What would you do if it can't be converted (e.g. I ask to convert time in ms to time in kg) throw a dimensionality error?

Yes, it should throw an error if the dimensionalities don't match. I imagine this will propagate naturally from the underlying functions. But should be one of the test cases.

cellmlmanip/model.py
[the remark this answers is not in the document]
skeating: So conclusion:

function is 
var = convert_variable(DummyVariable/ontology term/string name, Unit object, direction = {IN, OUT} enum)

returns the converted var - which may be same as original argument if no conversion was necessary
throws exceptions for
- Variable not in model
- Unit conversion impossible (ie m -> kg)

enum will be nicer - above just for clarity

we dont allow a string name for the unit

Everyone happy ??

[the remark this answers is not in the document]
jonc125: Yes to all of this, except that I'd suggest not allowing the 'string name' option for variable - too likely to cause confusion between whether a string is a URI (ontology term) or name.

cellmlmanip/model.py
```diff
@@ -9,14 +10,18 @@
...
 
-
 logger = logging.getLogger(__name__)
 
-
 # Delimiter for variables name in Sympy expressions: <component><delimiter><name>
 SYMPY_SYMBOL_DELIMITER = '$'
 
 
+class DataDirectionFlow(Enum):
```
jonc125: I think I suggested `DataFlowDirection` which makes more sense to me than `DataDirectionFlow`. But happy to hear arguments the other way :)

skeating: In fact to me `DataDirectionFlow` makes more sense but it doesn't really matter if you prefer the other. Maybe it's a male/female brain thing :-)

jonc125: OK, let's not bother changing it then :)

skeating:
Asking for review of this as it stands. With hope to merge with following questions:

1. Do I need to fiddle and get appveyor to pass

2. I could change current convert_variable so that the unit argument could also be a name/attributes pair which would allow it to be added if necessary while I work on #192

jonc125:
> Asking for review of this as it stands. With hope to merge with following questions:
> 
>     1. Do I need to fiddle and get appveyor to pass

No, sympy 1.5 support should be addressed in a separate PR.

>     2. I could change current convert_variable so that the unit argument could also be a name/attributes pair which would allow it to be added if necessary while I work on #192

I don't think we want to do this.

skeating:
Sorry @jonc125 just saw your post on slack. Code at present works if unit to be converted-to is in same registry as converted-from so it covers your suggestion. Weblab-fc would be place to take care of protocol/model using same registry and that I'll cover with taking over Michael's issue. I'll leave the other issue open and think a bit more.

cellmlmanip/model.py
```diff
@@ -619,6 +626,274 @@ def variables(self):
...
+        self.add_equation(expression)
+        return {'variable': new_deriv_variable, 'expression': eqn.args[0]}
+
+    def _convert_variable_instance(self, original_variable, cf, units, direction):
+        """
+        Internal function to create new variable and an equation for it.
+        :param original_variable: VariableDummy object to be converted [old units]
+        :param cf: conversion factor [new units/old units]
+        :param units: Unit object for new units
+        :param is_output: boolean are we adding input or output
```
jonc125: `is_output` has been replaced by `direction`

cellmlmanip/model.py
```diff
@@ -619,6 +626,274 @@ def variables(self):
...
+
+        # 3. copy cmeta_id from original and remove from original
+        new_variable = self.add_variable(name=new_name,
+                                         units=units,
+                                         initial_value=new_value,
+                                         cmeta_id=original_variable.cmeta_id)
+        original_variable.cmeta_id = ''
+
+        # 4. check whether we had an eqn for original
+        # if so, remove it and replace with new_var [new units] = rhs [old units] * cf [new units/old units]
```
jonc125: RHS if input, original_variable if output?

jonc125: Actually, this seems a slightly backwards way of doing the logic in sections 4 & 5. Would it be clearer to have at the outer level a switch on direction:

* If direction is OUTPUT, always just add an equation `new_variable = original_variable * cf`. No need to loop.
* If direction is INPUT, iterate over equations to find the original definition and replace with `new_variable = original_rhs * cf`, then add `original_variable = new_variable / cf`

Same effect overall, but the control is inverted and seems simpler to me. I had to read this three times to convince myself it worked.

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+    # fixtures
+
+    @pytest.fixture
+    def local_model(scope='function'):
+        """ Fixture to load a local copy of  the basic_ode model that may get modified. """
+        return shared.load_model('basic_ode')
+
+    @pytest.fixture
+    def model_missing_units(scope='function'):
+        """ Fixture to load a local copy of  the basic_ode model that may get modified. """
```
jonc125: These fixtures load different models...

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+    def multiode_freevar_model(scope='function'):
+        """ Fixture to load a local copy of  the basic_ode model that may get modified. """
+        return shared.load_model('repeated_ode_freevar_for_conversion_tests')
+
+    @pytest.fixture
+    def silly_names(scope='function'):
+        """ Fixture to load a local copy of  the basic_ode model that may get modified. """
+        return shared.load_model('silly_names')
+
+    def test_add_preferred_custom_unit_name(self, simple_ode_model):
```
jonc125: This isn't really a unit conversion test - should it be outside this class still?

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+            # cellml file states a (ms) = 1 (ms) + 1 (second)
+            bad_units_model.units.summarise_units(equation[0].rhs)
+
+        assert equation[1].lhs == symbol_b
+        with pytest.raises(units.UnitError):
+            # cellml file states b (per_ms) = power(a (ms), 1 (second))
+            bad_units_model.units.summarise_units(equation[1].rhs)
+
+    def test_add_input_state_variable(self, local_model):
+        """ Tests the Model.add_input function that changes units.
```
jonc125: The method isn't called add_input any more :)

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+            assert symbol_a in state_symbols
+            return True
+
+        mV_unit = local_model.get_units('mV')
+        volt_unit = local_model.get_units('volt')
+        original_var = local_model.get_symbol_by_name('env_ode$sv1')
+
+        assert test_original_state(local_model)
+        # test no change in units
+        newvar = local_model.convert_variable(original_var, mV_unit, DataDirectionFlow.INPUT)
```
jonc125: Probably worth doing this again with direction OUTPUT?

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+        assert len(local_model.variables()) == 5
+        symbol_a = local_model.get_symbol_by_cmeta_id('sv11')
+        assert local_model.get_initial_value(symbol_a) == 0.002
+        assert symbol_a.units == 'volt'
+        assert symbol_a.name == 'env_ode$sv1_converted'
+        symbol_t = local_model.get_symbol_by_cmeta_id('time')
+        assert symbol_t.units == 'ms'
+        symbol_orig = local_model.get_symbol_by_name('env_ode$sv1')
+        assert symbol_orig.units == 'mV'
+        assert local_model.get_initial_value(symbol_orig) == 2.0
```
jonc125: Hmm, we ought to delete the initial_value from the original variable, because it's now defined by an equation instead. Sorry I didn't catch that in the code above!

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+                var{time} time_converted: s;
+                var{sv11} sv1: mV {init: 2};
+                var sv1_orig_deriv mV_per_ms
+
+                time = 1000 * time_converted;
+                sv1_orig_deriv = 1{mV_per_ms}
+                ode(sv1, time_converted) = 1000 * sv1_orig_deriv
+        """
+        # original state
+        def test_original_state(local_model):
```
jonc125: Since this is essentially the same as with the tests above, perhaps extract as a method - it just needs a name that doesn't start with `test_`, e.g. `_check_original_state`?

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+        assert symbol_orig.units == 'mV'
+        symbol_derv = local_model.get_symbol_by_name('env_ode$sv1_orig_deriv')
+        assert symbol_derv.units == 'mV / ms'
+        assert len(local_model.equations) == 3
+        assert str(local_model.equations[0]) == 'Eq(_environment$time, 1000.0*_environment$time_converted)'
+        assert str(local_model.equations[1]) == 'Eq(_env_ode$sv1_orig_deriv, _1.0)'
+        assert str(local_model.equations[2]) == 'Eq(Derivative(_env_ode$sv1, _environment$time_converted), ' \
+                                                '1000.0*_env_ode$sv1_orig_deriv)'
+
+    def test_add_input_free_variable_ontology_term(self, local_model):
```
jonc125: If we get rid of the support for a string input, it cuts our tests in half :)

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+                ode(y, time) = 2{mV_per_ms}
+
+        convert_variable(time, second, DataDirectionFlow.INPUT)
+
+            becomes
+                var time: ms;
+                var{time} time_converted: s;
+                var{sv11} sv1: mV {init: 2};
+                var sv1_orig_deriv mV_per_ms
+                var y : mV
```
jonc125: y keeps the `{init: 3}`

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+        nA_unit = literals_model.get_units('nA')
+        original_var = literals_model.get_symbol_by_name('env_ode$x')
+
+        assert test_original_state(literals_model)
+        # test no change in units
+        literals_model.convert_variable(original_var, pA_unit, DataDirectionFlow.OUTPUT)
+        assert test_original_state(literals_model)
+
+        # change pA to nA
+        literals_model.convert_variable(original_var, nA_unit, DataDirectionFlow.OUTPUT)
```
jonc125: Is it worth checking the return value here?

tests/test_unit_conversion.py
jonc125: Ah, I see, this is tested in other tests below instead.

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+
+        # ontology term not present in model
+        with pytest.raises(AssertionError):
+            local_model.convert_variable('current', unit, direction)
+
+        # unit conversion is impossible
+        with pytest.raises(DimensionalityError):
+            local_model.convert_variable(variable, bad_unit, direction)
+
+    def test_noconversion_necessary(self, local_model):
```
jonc125: I don't think we need this, as it's done in all the tests above.

tests/test_unit_conversion.py
```diff
@@ -1,80 +1,1235 @@
...
+    #         var y: per_pA;
+    #         var{current} x_converted: nA
+    #
+    #         ode(sv1, time) = 1 :mV_per_ms;
+    #         x = 1000 * x_converted: pA;
+    #         y = 1{dimensionless}/x;
+    #         x_converted = 0.001 * 1 : nA
+    #     """
+    #     pA_unit = model_missing_units.get_units('pA')
+    #     nA_unit = literals_model.get_units('nA')
```
jonc125: So if we want to test this, I'd instead do something like
```
nA_unit = model_missing_units.units.ureg('nanoampere')
```
and not use the `literals_model` fixture.

If we get rid of the test, then remember to remove the fixture & CellML file.

cellmlmanip/model.py
```diff
@@ -619,6 +626,274 @@ def variables(self):
...
+                          equations will be adjusted
+                          or DataDirectionFlow.OUTPUT; the variable to be changed is an output, equations
+                          are unaffected apart from converting the actual output
+        :return: the original variable but with new units (or original unchanged if conversion was not necessary
+                 or impossible
+        :throws: AssertionError if the arguments are of incorrect type
+                                   the variable does not exist in the model
+                DimensionalityError if the unit conversion is impossible
+        """
+        original_variable = self._check_arguments(variable, units, direction)
```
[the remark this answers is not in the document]
skeating: Not sure what you mean by this.  Removing string as variable option means that this function does just check type and no longer returns a variable.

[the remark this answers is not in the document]
jonc125: I meant that 'check_arguments' is quite a generic kind of name - other methods might need to check their arguments in a different way. So naming the method 'check_conversion_arguments' makes it clearer what methods it supports.

tests/test_unit_conversion.py
skeating: Yes, because I need different models AND I am going to change them so they need local scope. Using the function scope does seem to mean I get a fresh version each function. I could try and make one model that got me all the examples I need but because I check for no change as well as if there is - this impacts all my tests so I went with multiple fixtures. Is this an issue - it hasnt given me any problems ?

jonc125: Not an issue at all - I was merely commenting on the mismatch between the docstring and fixture name. Sorry for being unclear!

tests/test_unit_conversion.py
skeating: unless its an output (or unless there are other problems in my code !)

jonc125: Indeed!

cellmlmanip/model.py
```diff
@@ -619,6 +626,276 @@ def variables(self):
...
+
+        # 2. if original has initial_value calculate new initial value
+        new_value = None
+        if original_variable.initial_value:
+            new_value = original_variable.initial_value * cf
+
+        # 3. copy cmeta_id from original and remove from original
+        new_variable = self.add_variable(name=new_name,
+                                         units=units,
+                                         initial_value=new_value,
```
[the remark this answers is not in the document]
MichaelClerx: Hey @jonc125 did we ever address this one?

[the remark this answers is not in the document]
MichaelClerx: Yes

PR #181, opened 2019-12-10
Added RNG validation for CellML & MathML

### LINKED ISSUE #10: Validate MathML input
Note from Jonathan:
> This would be a separate module, probably just validating against a (RELAX NG) schema. There's one in the old pycml.

MichaelClerx:
See #10 

This adds RNG validation (for CellML and MathML) and an invalid CellML file to test it.


cellmlmanip/parser.py
```diff
@@ -406,3 +415,21 @@ def _parent_of(parent_name, child_name):
...
+
+    def _validate(self, parser, tree):
+        """
+        Validates the given lxml ``tree`` against the CellML 1.0 RELAX NG schema.
+
+        :param parser: An `lxml.etree.XMLParser`
+        :param tree: An `lxml.etree.ElementTree` made with the given parser.
+        """
+
+        # Create RelaxNG object
```
jonc125: If a parser might be used for several models, should this be instantiated in the constructor and saved? Very minor point though!

cellmlmanip/data/cellml_1_0.rng
```diff
@@ -0,0 +1,989 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
jonc125: Should we include the .rnc source for this too, for completeness?

cellmlmanip/parser.py
MichaelClerx: Regarding that: What do you think of parser objects, that are mostly single-use? I always design them as classes but only because it groups my code nicely and I feel a bit dirty writing a bunch of private methods in module scope. I looked around a bit and parsers are usually classes (with non-static methods), although example code only uses them once.

Off-topic I know :D

MichaelClerx: Actually, that is relevant here because this parser class can only be used once :D

cellmlmanip/data/cellml_1_0.rng
MichaelClerx: Just in case someone wants to extract it?

MichaelClerx: Sure!

cellmlmanip/parser.py
jonc125: I normally make them classes too. A class that only gets used once is still fine! Of course in some languages everything is a class...

cellmlmanip/data/cellml_1_0.rng
MichaelClerx: I've also stuck them here, by the way: https://github.com/MichaelClerx/cellml-validation/tree/master/cellml_1_0
Hope you don't mind!
([it works well but gets beaten by OpenCOR](https://github.com/MichaelClerx/cellml-validation/tree/master/reports))

jonc125: Nice :)

PR #186, opened 2019-12-12
implemented method to get derived quantities

### LINKED ISSUE #185: Determine derived quantities
I need to be able to determine derived quantities for chaste codegen.

@maurice I was replying to your last comment when it disappeared!
Anyway, you can use `get_symbols_by_rdf` to find things annotated explicitly as d-q and also everything with an oxmeta name. Then use something like:
[v for v, node in self.graph.nodes.items() if node.get('variable_type', '') not in ('state', 'free', 'parameter') and v in oxmeta_symbols]

jonathan  27 minutes ago
You could write that in -cg, or could add a method to cellmlmanip. Although the v in oxmeta_symbols bit probably wouldn't be appropriate for cellmlmanip

jonathan  27 minutes ago
But getting all the derived quantities could be a method, like for state vars etc.

MauriceHendrix:
## Description
Added a function to the model to be able to get derived quantities.

## Motivation and Context

Fixes #185 
Chaste codegen needs to be able to determine derived quantities, and it may be useful for others to have in cellmlmanip

## Types of changes

## Checklist:

- [X] I have checked spelling in (new) comments.

## Testing


cellmlmanip/model.py
```diff
@@ -267,6 +267,17 @@ def get_derivative_symbols(self):
         derivative_symbols = [v for v in self.graph if isinstance(v, sympy.Derivative)]
         return sorted(derivative_symbols, key=lambda state_var: state_var.args[0].order_added)
 
+    def get_derived_quantities(self, namespace_uri=None):
+        """Returns a list of derived quantities found in the given model graph.
+        A derived quantity is anything is not a state variable, free variable or parameter.
```
MichaelClerx: ...any variable that is...

MichaelClerx: ... parameter/constant.

Would that make it clearer for the user?

cellmlmanip/model.py
```diff
@@ -267,6 +267,17 @@ def get_derivative_symbols(self):
...
+
+        :param namespace_uri only return variables with ``{namespace_uri}annotation_name`` annotation are returned
```
MichaelClerx: Two times return in this sentence!

tests/test_model_functions.py
```diff
@@ -62,6 +62,24 @@ def test_graph_for_dae(self):
     #######################################################################
     # this section contains tests for each get_XXX function on Model
 
+    def test_get_derived_quantities(self, basic_model):
+        """ Tests Model.get_state_symbols() works correctly. """
+
+        derived_quantities = basic_model.get_derived_quantities()
+        assert str(derived_quantities) == '[Derivative(_env_ode$sv1, _environment$time)]'
```
MichaelClerx: I thought this would be excluded as it's  a state?

MauriceHendrix: `_environment$time` is but the derivative isn't it is a deriv of course.

MauriceHendrix:
> Are you sure that a derivative counts as a derived quantity? I thought those were only things like INa etc. ?

apparently you're right so I'll fix that.

cellmlmanip/model.py
```diff
@@ -267,6 +267,18 @@ def get_derivative_symbols(self):
         derivative_symbols = [v for v in self.graph if isinstance(v, sympy.Derivative)]
         return sorted(derivative_symbols, key=lambda state_var: state_var.args[0].order_added)
 
+    def get_derived_quantities(self, namespace_uri=None):
```
jonc125: I think rather than putting the namespace option in this method, I'd prefer the calling code to filter the returned list. It'll make this method simpler, and doesn't really add much complexity for the caller.

tests/test_model_functions.py
```diff
@@ -62,6 +62,35 @@ def test_graph_for_dae(self):
...
+        derived_quantities = aslanidi_model.get_derived_quantities()
+        assert len(derived_quantities) == 78
+        derived_quantities = aslanidi_model.get_derived_quantities(namespace_uri=shared.OXMETA)
+        assert str(derived_quantities) == '[_membrane$i_Stim]'
+        derived_quantities = aslanidi_model.get_derived_quantities(namespace_uri='http://example.com#')
+        assert len(derived_quantities) == 0
+
+    def test_get_derived_quantities3(self):
+        """ Tests Model.get_state_symbols() works correctly. """
+        model = shared.load_model('matsuoka_model_2003')
```
jonc125: I'd rather not add another large real model just to test this! If we need to, we can add to the simple test models. But my comment above may make this unnecessary.

This test feels like it's checking how CG wants to use the method, rather than checking the method itself. Which is probably better tested in CG.

tests/test_model_functions.py
```diff
@@ -62,6 +62,35 @@ def test_graph_for_dae(self):
     #######################################################################
     # this section contains tests for each get_XXX function on Model
 
+    def test_get_derived_quantities(self, basic_model):
```
jonc125: We could also check `test_simple_odes` which does have at least one derived quantity (circle_y_implementation$rhs).

tests/test_model_functions.py
MauriceHendrix: yes of course I'll remove it once it works, i'm just asking why it's not working as expected????

tests/test_model_functions.py
MauriceHendrix: ok

MauriceHendrix: seems to have 12 (see test) is that what we're expecting?

tests/test_model_functions.py
MauriceHendrix: it's gone now

tests/test_model_functions.py
jonc125: Yes, all the ones it picked up look OK to me - including the two that arise from units-converted connections!

PR #190, opened 2019-12-16
Stop simplify

MauriceHendrix:
## Description

I know this fails due to the follwoing, which is happening for a couple of models, but why?
```
        # What's going on here?
        model = shared.load_model('viswanathan_model_1999_epi')
        membrane_stimulus_current = model.get_symbol_by_ontology_term(shared.OXMETA, 'membrane_stimulus_current')
        eqs = model.get_equations_for([membrane_stimulus_current])
```

On line 256 in test_model_functions.py (the ther errors go away if i take that out)

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -173,7 +173,8 @@ class Model(object):
             target.assigned_to = source.assigned_to
             # everywhere the target variable is used, replace with source variable
             for index, equation in enumerate(self.equations):
-                self.equations[index] = equation.xreplace({target: source.assigned_to})
+                with sympy.evaluate(False):  # prevent simplifying now, as that may introduce float precision errors
+                    self.equations[index] = equation.xreplace({target: source.assigned_to})
 
         # Otherwise, this connection requires a conversion
         else:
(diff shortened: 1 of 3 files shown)


MichaelClerx:
I don't think this functionality is ready:

https://github.com/sympy/sympy/blob/master/sympy/core/evaluate.py#L25-L27

    @contextmanager
    def evaluate(x):
        """ Control automatic evaluation

        This context manager controls whether or not all SymPy functions evaluate
        by default.

        Note that much of SymPy expects evaluated expressions.  This functionality
        is experimental and is unlikely to function as intended on large
        expressions.

MauriceHendrix:
So basically we're *£@#&-+()
Seems odd that not simplifying should be experimental

jonc125:
I think we need to read through https://docs.sympy.org/latest/modules/codegen.html, especially the bits on cfunctions and rewriting!

MauriceHendrix:
> I think we need to read through https://docs.sympy.org/latest/modules/codegen.html, especially the bits on cfunctions and rewriting!

Interesting but we already have a printer, that can do this as well as control our variable names

jonc125:
I didn't mean to switch to their printer. But the cfunctions and rewriting stuff might still be useful. For instance we could use things like `expm1` to optimise how some numeric edge cases are handled in the generated code. And indeed update the transpiler to use the log10 function! I'll open issues for these.

PR #194, opened 2020-01-07
simplied get_ontology_terms_by_symbol to use symbol.cmeta_id

MauriceHendrix:
## Description

in `get_ontology_terms_by_symbol` it now checks and uses `symbol.cmeta_id` instead of `self.graph.nodes[symbol].get('cmeta_id', None)`

## Motivation and Context

This is simpler and it prevents issues with variables that are not in the graph (e.g. because they aren't involved in any equations)

## Types of changes

## Checklist:

- [ ] I have updated all documentation necessary.
No comments changed

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -404,13 +404,12 @@ class Model(object):
         if symbol.cmeta_id:
             predicate = ('http://biomodels.net/biology-qualifiers/', 'is')
             predicate = create_rdf_node(predicate)
-            for delimeter in ('#', '/'):  # Look for terms using either possible namespace delimiter
-                subject = rdflib.term.URIRef(delimeter + symbol.cmeta_id)
-                for object in self.rdf.objects(subject, predicate):
-                    # We are only interested in annotation within the namespace
-                    if namespace_uri is None or str(object).startswith(namespace_uri):
-                        uri_parts = str(object).split(delimeter)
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -401,12 +401,11 @@ def get_ontology_terms_by_symbol(self, symbol, namespace_uri=None):
...
         """
         ontology_terms = []
-        cmeta_id = self.graph.nodes[symbol].get('cmeta_id', None)
-        if cmeta_id:
+        if symbol.cmeta_id:
             predicate = ('http://biomodels.net/biology-qualifiers/', 'is')
             predicate = create_rdf_node(predicate)
             for delimeter in ('#', '/'):  # Look for terms using either possible namespace delimiter
-                subject = rdflib.term.URIRef(delimeter + cmeta_id)
+                subject = rdflib.term.URIRef(delimeter + symbol.cmeta_id)
```
jonc125: Ah, not something you introduced, but this should always be `'#' + symbol.cmeta_id`!

MauriceHendrix: Are you absolutely sure about that? I distinctly remember `/` being possible too.

jonc125: (You can have # or / between NS URI and local part, but not as a pure prefix for a within-document id.)

PR #196, opened 2020-01-13
Conversion same units different name

### LINKED ISSUE #197: cellmlmanip conversion code converts variables that are the same units but have different names
A thing I noticed about the conversion code: It checks if the units are the same to avoid needlessly converting. Great!
However I have a case where I convert from mM to millimolar and the result gives me 1.0 * var_chaste_interface__Ca__Ca_i_converted; // mM
Is this on purpose? (after conversion you definitely know for sure the unit is millimolar so that might be useful for some)

jonathan  4 hours ago
I doubt it's on purpose; it might be to do with how the 'same units' check works. I can't recall if it checks they're the same object, or something else

maurice  4 hours ago
it does indeed

maurice  3 hours ago
just wanted o raise it before i propose a change. On th one hand I'd not expect it to convert but on the other hand I'd also expect the units to be exactly what I told it to convert to.
but if I have 2 variables in the same unit say mM and I convert one to say millimolar it would also surprise me if they're suddenly not the same unit

jonathan  3 hours ago
I think it shouldn't convert in this situation

jonathan  2 hours ago
Hmm, though it uses == rather than is, so I'd have expected it to avoid a conversion here! Clearly there's something in the internal representation of each that causes a difference.

Sarah Keating  2 hours ago
The issue is also with unit registry because you can use 'string' representation for units. It does not consider mM == millimolar. It will add both to the registry and if you step through the == code the strings get compared.  So we need a custom units== function that compares everything except the 'name string'

Sarah Keating  2 hours ago
UNLESS you used
add_preferred_custom_unit_name
in your code and then it 'changes' all the same units to use the same name (edited) 

maurice  1 hour ago
I thought we had removed add_preferred_custom_unit_name ? I stopped using it because it was pretty slow

Sarah Keating  1 hour ago
its still there

maurice  1 hour ago
@jonathan would it make sense to remove add_preferred_custom_unit_name ?
:+1:

Sarah Keating  34 minutes ago
typo

jonathan  5 minutes ago
Unless anything else is using it, yes

jonathan  16 minutes ago
I'd be wary of suppressing errors (especially as it makes the cellmlmanip code uglier) - you'd probably want at least to be warning the user in such cases, rather than doing no conversion silently

MauriceHendrix:
## Description

uses conversion_factor==1.0 to determine if units are the same.

## Motivation and Context

fixes #197  units that are the same except their string name will get seen as the same (and not converted)

I also added an optional flag to not get errors thrown if a unit conversion isn't possible. In that case it just returns the original variable. the reason is that I have quite a few cases where I want to try and convert but it may not always be possible (e.g. sometimes calcium_concentration has the units dimensionless). Personally i find it easier if I don't need to write elaborate checks but that's open for discussion I suppose.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -689,7 +689,7 @@ class Model(object):
 
         original_units = original_variable.units
         # no conversion necessary
-        if self.units.get_conversion_factor(units, from_unit=original_units) == 1.0:
+        if  1 * original_units == 1 * units:
             return original_variable
 
         # conversion_factor for old units to new
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -689,7 +689,7 @@ class Model(object):
 
         original_units = original_variable.units
         # no conversion necessary
-        if  1 * original_units == 1 * units:
+        if 1 * original_units == 1 * units:
             return original_variable
 
         # conversion_factor for old units to new
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -624,7 +624,7 @@ class Model(object):
         """ Returns an iterator over this model's variable symbols. """
         return self._name_to_symbol.values()
 
-    def convert_variable(self, original_variable, units, direction, raise_errors=True):
+    def convert_variable(self, original_variable, units, direction):
         """
         Add a new linked version of the given variable in the desired units.
 
@@ -679,19 +679,13 @@ class Model(object):
                           equations will be adjusted
                           or DataDirectionFlow.OUTPUT; the variable to be changed is an output, equations
                           are unaffected apart from converting the actual output
-        :param raise_errors: determines whether the method will throw an error if one of the checks fails
-                          if raise_errors==False and checks fail, the original variable is returned unchanged
         :return: new variable with desired units, or original unchanged if conversion was not necessary
         :throws: AssertionError if the arguments are of incorrect type
                                 or the variable does not exist in the model
                 DimensionalityError if the unit conversion is impossible
         """
         # assertion errors will be thrown here if arguments are incorrect type
-        checks_passed = self._check_arguments_for_convert_variables(original_variable, units, direction, raise_errors)
-
-        # no conversion possible (checks failed)
-        if not checks_passed:
-            return original_variable
+        self._check_arguments_for_convert_variables(original_variable, units, direction)
 
         original_units = original_variable.units
         # no conversion necessary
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -679,17 +679,23 @@ def convert_variable(self, original_variable, units, direction):
...
+        checks_passed = self._check_arguments_for_convert_variables(original_variable, units, direction, raise_errors)
+
+        # no conversion possible (checks failed)
+        if not checks_passed:
+            return original_variable
 
         original_units = original_variable.units
         # no conversion necessary
-        if original_units == units:
+        if self.units.get_conversion_factor(units, from_unit=original_units) == 1.0:
```
jonc125: I think I've found a simpler way to do this:
```
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> ureg.define('my_mM = 0.001 * molar')
>>> ureg.my_mM == ureg.mM
False
>>> 1 * ureg.my_mM == 1 * ureg.mM
True
```
So we can just do here:
```python
if 1 * original_units == 1 * units:  # Still compares equal if the unit names differ
```

cellmlmanip/model.py
```diff
@@ -624,7 +624,7 @@ def variables(self):
         """ Returns an iterator over this model's variable symbols. """
         return self._name_to_symbol.values()
 
-    def convert_variable(self, original_variable, units, direction):
+    def convert_variable(self, original_variable, units, direction, raise_errors=True):
```
jonc125: As mentioned on Slack I'm not keen on this change.

tests/test_unit_conversion.py
```diff
@@ -711,26 +710,32 @@ def test_convert_variable_invalid_arguments(self, local_model):
         # arguments wrong types
         with pytest.raises(AssertionError):
             local_model.convert_variable('x', unit, direction)
+        local_model.convert_variable('x', unit, direction, raise_errors=False) == 'x'
```
jonc125: There are asserts missing here. But in any case I don't think we want this feature.

cellmlmanip/model.py
MauriceHendrix: ok

jonc125:
Ah, approved except for flake8 issues!

MauriceHendrix:
> Ah, approved except for flake8 issues!
whuups I did check but in the wrong window :)

Issue #214, opened 2020-02-04
Allow UnitStore to share a registry with an existing unit store


MichaelClerx:
- [ ] So that the pint registry is the same, but the names are completely separated
- [ ] When parsing a model it should be possible to pass in an existing unit store

MichaelClerx:
It seems the unit registry is responsible for maintaining the names. So for example if we create a unit `mm = 1000 * meter` then `mm` gets added to the registry. This means that later we can ask the registry to parse strings like `5 * mm`. So having a shared registry would mean we can't separate unit namespaces.

MichaelClerx:
So now think it'd be better to just have distinct unit registries, and to just use quantities everywhere (comparing quantities from different registries is allowed)

```
#!/usr/bin/env python3
import cellmlmanip
import cellmlmanip.units

import os

import pint

path = os.path.join('cellmlmanip', 'data', 'cellml_units.txt')
r1 = pint.UnitRegistry(path)
r2 = pint.UnitRegistry(path)

c1 = r1.candela
c2 = r2.candela

print(c1, type(c1))
print(c2, type(c2))

#print(c1 == c2)

q1 = 1 * c1
q2 = 1 * c2

print(q1 == q2)
```

jonc125:
That sounds like a workable plan!

MichaelClerx:
I thought so, but seems `pint` isn't set up for that:

```
#!/usr/bin/env python3
import cellmlmanip
import cellmlmanip.units

import os

import pint

path = os.path.join('cellmlmanip', 'data', 'cellml_units.txt')
r1 = pint.UnitRegistry(path)
r2 = pint.UnitRegistry(path)

r1.define('abc=meter*0.2')
r2.define('xyz=meter*0.2')

u1 = r1.abc
u2 = r2.xyz

print(u1, type(u1))
print(u2, type(u2))

q1 = 1 * u1
q2 = 1 * u2

print(q1, type(q1))
print(q2, type(q2))

print(q1 == q2)
```

jonc125:
Drat! Do we need to share a registry after all, and just make `UnitStore` prefix unit names with a uuid say before defining them in pint's registry?

MichaelClerx:
Yeah if we run a regex over each definition the user passes in to turn e.g.
``mm = meter * 1000`` into ``special_mm = special_meter * 1000``

This will make error output harder to read though, as the user will see all the weird names

PR #215, opened 2020-02-04
Allow multiple UnitStore objects to share the same pint unit registry

MichaelClerx:
## Description
Originally meant to do #214 
This is now more of a unit tidying / refactoring PR.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -492,12 +492,6 @@ class Model(object):
                     ontology_terms.append(uri_parts[-1])
         return ontology_terms
 
-    def get_units(self, name):
-        """
-        Looks up and returns a pint `Unit` object with the given name.
-        """
-        return self.units.get_unit(name)
-
     def get_definition(self, symbol):
         """Get the equation (if any) defining the given variable.
 
@@ -509,6 +503,12 @@ class Model(object):
         if defn is None:
             defn = self._var_definition_map.get(symbol)
         return defn
+
+    def get_units(self, name):
+        """
+        Looks up and returns a pint `Unit` object with the given name.
+        """
+        return self.units.get_unit(name)
 
     @property
     def graph(self):
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -486,12 +486,6 @@ class Model(object):
             defn = self._var_definition_map.get(symbol)
         return defn
 
-    def get_units(self, name):
-        """
-        Looks up and returns a pint `Unit` object with the given name.
-        """
-        return self.units.get_unit(name)
-
     @property
     def graph(self):
         """ A ``networkx.DiGraph`` containing the model equations. """
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -259,7 +259,7 @@ class Model(object):
         # Otherwise, this connection requires a conversion
         else:
             # Get the scaling factor required to convert source units to target units
-            factor = self.units.convert_to(1 * source.units, target.units).magnitude
+            factor = self.units.convert(1 * source.units, target.units).magnitude
 
             # Dummy to represent this factor in equations, having units for conversion
             factor_dummy = self.add_number(factor, target.units / source.units)
@@ -290,7 +290,7 @@ class Model(object):
         """
         lhs_units = self.units.evaluate_units(equality.lhs)
         rhs_units = self.units.evaluate_units(equality.rhs)
-        assert self.units.is_equal(rhs_units, lhs_units), 'Units %s %s != %s %s' % (
+        assert self.units.is_equivalent(rhs_units, lhs_units), 'Units %s %s != %s %s' % (
             lhs_units, self.units.show_base_units(lhs_units),
             rhs_units, self.units.show_base_units(rhs_units)
         )
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -290,13 +290,11 @@ class Model(object):
         Checks whether the LHS and RHS in a ``sympy.Eq`` have the same units.
         :param equality: A ``sympy.Eq``.
         """
-        lhs_units = self.units.summarise_units(equality.lhs)
-        rhs_units = self.units.summarise_units(equality.rhs)
-
-        # TODO REWRITE WITHOUT USING PRIVATE PROPERTIES
-        assert self.units.is_unit_equal(rhs_units, lhs_units), 'Units %s %s != %s %s' % (
-            lhs_units, self.units._registry.get_base_units(lhs_units),
-            rhs_units, self.units._registry.get_base_units(rhs_units)
+        lhs_units = self.units.evaluate_units(equality.lhs)
+        rhs_units = self.units.evaluate_units(equality.rhs)
+        assert self.units.is_equal(rhs_units, lhs_units), 'Units %s %s != %s %s' % (
+            lhs_units, self.units.show_base_units(lhs_units),
+            rhs_units, self.units.show_base_units(rhs_units)
         )
 
     def get_equations_for(self, symbols, recurse=True, strip_units=True):
@@ -847,10 +845,8 @@ class Model(object):
         """
         # 1. create a new variable
         deriv_name = self._get_unique_name(derivative_variable.name + '_orig_deriv')
-        # TODO REWRITE THIS WITHOUT USING PRIVATE PROPERTIES
-        deriv_units = self.units._calculator.traverse(eqn.args[0])
-        new_deriv_variable = self.add_variable(name=deriv_name,
-                                               units=deriv_units.units)
+        deriv_units = self.units.evaluate_units(eqn.args[0])
+        new_deriv_variable = self.add_variable(name=deriv_name, units=deriv_units)
 
         # 2. create new equation and remove original
         expression = sympy.Eq(new_deriv_variable, eqn.args[1])
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -74,24 +74,6 @@ class Model(object):
         # Map from VariableDummy to defining equation, where the variable is defined by an ODE
         self._ode_definition_map = {}
 
-    '''
-    def add_unit(self, name, expression):
-        """
-        Adds a unit of measurement to this model, with a given ``name`` and list of ``attributes``.
-
-        :param name: A string name.
-        :param attributes: An optional list of dictionaries containing unit attributes. See
-            :meth:`UnitStore.add_custom_unit()`.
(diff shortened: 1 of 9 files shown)


MichaelClerx:
```
        # Without shared registry
        store1 = UnitStore()
        store2 = UnitStore()
        assert not store1.is_unit_defined('mmm')
        assert not store2.is_unit_defined('mmm')

        # ...adding unit to store 1 doesn't affect store 2
        unit_attributes = [{'prefix': 'milli', 'units': 'metre'}]
        store1.add_custom_unit('mmm', unit_attributes)
        assert store1.is_unit_defined('mmm')
        assert not store2.is_unit_defined('mmm')
        store2.add_custom_unit('mmm', unit_attributes)
        assert store2.is_unit_defined('mmm')

        # ...and the units do not count as equivalent, because they're from different registries (even thought they have
        # the same basi SI unit expansion).
        with pytest.raises(ValueError):
            assert store1.ureg.mmm != store2.ureg.mmm

        # ...but quantities are the same
        assert store1.ureg.parse_expression('mmm') == store2.ureg.parse_expression('mmm')

        # With a shared registry
        store2 = UnitStore(store1.ureg)
        assert store1.is_unit_defined('mmm')
        assert not store2.is_unit_defined('mmm')
```

Final test fails here, because that method just asks the registry for a unit (and that has it!). So guessing I need to do a bit more rewriting work here to stop the UnitStore from exposing all units in the underlying registry? (So only CellML units and custom added units can be seen)? @jonc125

jonc125:
> Final test fails here, because that method just asks the registry for a unit (and that has it!). So guessing I need to do a bit more rewriting work here to stop the UnitStore from exposing all units in the underlying registry? (So only CellML units and custom added units can be seen)? @jonc125

Yes, I think so - basically anything looking up names / checking names needs to check the `UnitStore` list not the registry's.

MichaelClerx:
@jonc125 could you have an early look at this?

Changes include:

- refactoring so that only the parser knows about CellML unit attributes
- instead we can add units with e.g. add_units('mm', 'meter / 1000')
- hiding unit internals from the user (this is prettier, but also needed if we're going to prefix all the unit names with some uid)

cellmlmanip/model.py
```diff
@@ -69,8 +69,11 @@ def __init__(self, name, cmeta_id=None):
         # Map from VariableDummy to defining equation, where the variable is defined by an ODE
         self._ode_definition_map = {}
 
-    def add_unit(self, name, attributes=None, base_units=False):
+    '''
+    def add_unit(self, name, expression):
         """
+        Ass a
```
jonc125: oops ;)

cellmlmanip/model.py
```diff
@@ -825,7 +832,8 @@ def _create_new_deriv_variable_and_equation(self, eqn, derivative_variable):
         """
         # 1. create a new variable
         deriv_name = self._get_unique_name(derivative_variable.name + '_orig_deriv')
-        deriv_units = self.units.calculator.traverse(eqn.args[0])
+        # TODO REWRITE THIS WITHOUT USING PRIVATE PROPERTIES
```
jonc125: Why does the calculator need to be private?

tests/test_units.py
```diff
@@ -20,203 +20,221 @@
...
-            quantity_store.get_quantity('kilogram'),
-            unit_registry.kilogram
-        )
+            assert unit_store.is_unit_defined(name)
 
         # Custom units defined in CellML example
-        assert quantity_store.is_unit_equal(
-            quantity_store.get_quantity('per_ms'),
-            unit_registry.millisecond**(-1)
+        '''
```
jonc125: Are you meaning to skip these?

tests/test_units.py
```diff
@@ -20,203 +20,221 @@
...
         """ Tests Units.get_conversion_factor() function. """
-        ureg = quantity_store.ureg
-        assert quantity_store.get_conversion_factor(quantity=1 * ureg.ms, to_unit=ureg.second) == 0.001
-        assert quantity_store.get_conversion_factor(quantity=1 * ureg.volt, to_unit=ureg.mV) == 1000.0
-
-        assert quantity_store.get_conversion_factor(
-            quantity=1 * quantity_store.get_quantity('milli_mole'),
-            to_unit=quantity_store.get_quantity('mole')
+        assert unit_store.get_conversion_factor(
+            quantity=1 * unit_store.get_quantity('ms'),
```
jonc125: Given the method name is `get_quantity` do we need the `1 * `? Or is the method really `get_unit`?

Should the `to_unit` argument on the next line be changed?

tests/test_units.py
```diff
@@ -20,203 +20,221 @@
...
+    # def test_make_pint_unit_definition(self, unit_store):
+    #    """ Tests Units._make_pint_unit_definition() function. """
+    #    unit_attributes = [{'prefix': -3, 'units': 'metre'},
+    #                       {'prefix': 'milli', 'exponent': -1, 'units': 'second'},
+    #                       {'multiplier': 2, 'units': 'kilograms'}
+    #                       ]
+    #    assert(unit_store._make_pint_unit_definition('kg_mm_per_ms', unit_attributes) ==
+    #           'kg_mm_per_ms=(meter * 1e-3)*(((second * 0.001))**-1)*(2 * kilogram)')
+
+    '''
```
jonc125: I'm guessing this bit is just work in progress.

cellmlmanip/units.py
```diff
@@ -162,146 +143,145 @@ def __init__(self, expression):
...
+        if store is None:
+            # Create new unit registry, configured for CellML units.
+            self._registry = pint.UnitRegistry(os.path.join(os.path.dirname(__file__), 'data', 'cellml_units.txt'))
+        else:
+            # Share a registry and calculator with the given unit store
+            self._registry = store._registry
 
-        # Unit calculator
-        self.calculator = UnitCalculator(self.ureg)
+        # User defined units
```
jonc125: I'd note here that the set just stores names. I thought initially it contained the objects.

cellmlmanip/units.py
```diff
@@ -162,146 +143,145 @@ def __init__(self, expression):
...
-        self.custom_defined.add(units_name)
-
     def get_quantity(self, unit_name):
         """Returns a pint `Unit` with the given name from the UnitRegistry.
 
         :param unit_name: string name of the unit
         :returns: `Unit`
         :raises: ``pint.UndefinedUnitError`` if the unit is not present in registry.
         """
+        # TODO CHANGE NAME TO GET UNIT, IF IT RETURNS A UNIT, ELSE UPDATE DOCSTRING
```
jonc125: Hah, you've already thought of my earlier comment :)

cellmlmanip/units.py
```diff
@@ -162,146 +143,145 @@ def __init__(self, expression):
...
+        """Check if a unit with the given ``name`` exists.
 
         :param name: string name of the unit
         :returns: True if exists, else False
         """
+        # TODO ADD PREFIX
+
         try:
-            getattr(self.ureg, name)
+            getattr(self._registry, name)
```
jonc125: I was thinking we could use (a variant of) `self._custom_units` for this, if it also contained the built-in CellML names?

cellmlmanip/units.py
```diff
@@ -162,146 +143,145 @@ def __init__(self, expression):
...
-        full_unit_expr = '*'.join(full_unit_expr)
-
-        # Return Pint definition string
-        logger.debug('Unit %s => %s', units_name, full_unit_expr)
-        return '%s=%s' % (units_name, full_unit_expr)
-
     def is_unit_equal(self, unit1, unit2):
-        """Compares whether two units are equivalent by converting them into base units and comparing the
-        resulting units and multiplicative factors
+        """Compares whether two units are equivalent by converting them into base units and comparing the resulting
```
jonc125: Once we've figured out the best way to compare units, we should make the implementation here match, and use it throughout the code (including CG) avoiding various 'is the conversion factor 1' type tests.

cellmlmanip/units.py
```diff
@@ -384,6 +366,8 @@ def dimensionally_equivalent(self, symbol1, symbol2):
         :param symbol2: the second expression to compare
         :returns: True if units are equal (regardless of quantity), False otherwise
         """
+        # TODO IF THEY ARE EXPRESSIONS THE ARGS SHOULD NOT BE CALLED SYMBOL
```
jonc125: Yep. Probably a legacy from before the calculator was created.

tests/test_units.py
MichaelClerx: If I can. They seem to be testing that `pint` works as intended. Happy to replace them if it turns out they're needed

cellmlmanip/model.py
MichaelClerx: As far as I can tell it's only a class (instead of a method) because it's long and has a few nested functions (now implemented as hidden methods of the class).

From a user's point of view it will be a lot simpler if you can just say e.g. `UnitStore.get_derived_units(expression)` and don't need to know that there's an exta class because it made the code prettier

tests/test_units.py
MichaelClerx: `get_quantity` should definitely be `get_unit`. Will check the rest later!

tests/test_units.py
MichaelClerx: I'm leaving these out unless they become public in the future. Testing should happen via the public API,

MichaelClerx:
@jonc125 shall we review & merge this one before I add the actual functionality? Should only be half an hour or so after this is in :D

cellmlmanip/units.py
```diff
@@ -162,271 +140,226 @@ def __init__(self, expression):
...
+        """Converts the given ``quantity`` to be in the given ``unit``.
 
-        # Return Pint definition string
-        logger.debug('Unit %s => %s', units_name, full_unit_expr)
-        return '%s=%s' % (units_name, full_unit_expr)
+        :param quantity: a ``Quantity`` to convert.
+        :param unit: a ``Unit`` to convert it to.
+        :returns: the converted ``Quantity``.
+        """
+        # TODO CHECK IF THESE ARE QUANTITIES OR UNITS
```
jonc125: todo done?

cellmlmanip/model.py
```diff
@@ -135,24 +121,24 @@ def add_number(self, value, units):
         Creates and returns a :class:`NumberDummy` to represent a number with units in sympy expressions.
 
         :param number: A number (anything convertible to float).
-        :param units: A `pint` units representation
+        :param units: A string representation of a unit or a ``Unit`` object.
```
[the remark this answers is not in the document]
MichaelClerx: Thanks. Updated it!

PR #220, opened 2020-02-06
Allow unit stores to share the same registry

MichaelClerx:
## Description
See #214 

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/parser.py
+++ b/cellmlmanip/parser.py
@@ -152,8 +152,8 @@ class Parser(object):
         units_elements = model.findall(Parser.with_ns(XmlNs.CELLML, 'units'))
 
         # get list of built-in cellml units
-        from cellmlmanip.units import CELLML_UNITS
-        units_found = set(CELLML_UNITS)
+        from cellmlmanip.units import _CELLML_UNITS
+        units_found = set(_CELLML_UNITS)
 
         # get all the units defined in the cellml model
         definitions_to_add = OrderedDict()
--- a/cellmlmanip/parser.py
+++ b/cellmlmanip/parser.py
@@ -13,7 +13,7 @@ from cellmlmanip.model import SYMPY_SYMBOL_DELIMITER, Model
 from cellmlmanip.transpiler import Transpiler
 
 
-_UNIT_PREFIXES = {
+UNIT_PREFIXES = {
     'yocto': 1e-24,
     'zepto': 1e-21,
     'atto': 1e-18,
@@ -212,18 +212,16 @@ class Parser(object):
 
         # For each of the <unit> elements for this unit definition
         for unit_element in unit_attributes:
-            # Source the <unit units="XXX"> from our store
-            matched_unit = self.model.units.get_unit(unit_element['units'])
 
-            # Construct a string representing the expression for this <unit>
-            expr = str(matched_unit)
+            # Start from the unit name
+            expr = unit_element['units']
 
             # See https://www.cellml.org/specifications/cellml_1.1/#sec_units 5.2.2
             # offset, prefix, exponent, and multiplier
 
             if 'prefix' in unit_element:
                 try:
-                    power = _UNIT_PREFIXES[unit_element['prefix']]
+                    power = UNIT_PREFIXES[unit_element['prefix']]
                 except KeyError:
                     # Assume that prefix is an integer.
                     power = '1e%s' % unit_element['prefix']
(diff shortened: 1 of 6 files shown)


tests/test_units.py
```diff
@@ -69,7 +77,7 @@ def test_get_unit(self):
 
         # Get CellML unit
         store = UnitStore()
-        assert str(store.get_unit('liter')) == 'liter'
+        assert str(store.get_unit('liter')).endswith('liter')
         assert isinstance(store.get_unit('ampere'), store.Unit)
 
         # Get user unit
```
jonc125: Other things it would be worth testing:
* If a user tries to `add_unit('store0_volt', 'millisecond')` or similar, do subsequent `get_unit` calls get very confused?
* If units named like 'my_unit' and 'also_my_unit' are added, does this work? Are both looked up correctly with `get_unit`?
* What about 'my_unit' and 'my_unit_again'?
* Or 'my_unit' and 'another_my_unit2'?

tests/test_units.py
```diff
@@ -69,7 +77,7 @@ def test_get_unit(self):
 
         # Get CellML unit
         store = UnitStore()
-        assert str(store.get_unit('liter')) == 'liter'
+        assert str(store.get_unit('liter')).endswith('liter')
```
jonc125: Except I thought CellML units didn't get prefixes, so this should still equal 'liter'?

cellmlmanip/parser.py
```diff
@@ -152,8 +152,8 @@ def _add_units(self, model):
         units_elements = model.findall(Parser.with_ns(XmlNs.CELLML, 'units'))
 
         # get list of built-in cellml units
-        from cellmlmanip.units import CELLML_UNITS
-        units_found = set(CELLML_UNITS)
+        from cellmlmanip.units import _CELLML_UNITS
```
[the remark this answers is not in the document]
MichaelClerx: It's a small point, I'm happy to change it if you like!
I tend to think of lists like these as implementation details rather than part of the public API. Not sure you could break things by modifying it, but you could certainly change the expected behaviour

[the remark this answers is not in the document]
jonc125: I was mainly thinking it might be useful to users as a read-only property. But actually you can get the info more usefully just by creating an empty `UnitStore`, so having this private makes more sense.

tests/test_units.py
```diff
@@ -29,6 +29,14 @@ def test_add_unit(self):
         unitstore.add_unit('u2', 'dimensionless * 3')
         assert unitstore.is_defined('u2')
 
+        # SI prefixes are allowed on user and CellML units
+        unitstore.add_unit('ms', 'milliu1')
+        unitstore.add_unit('mm', 'millimeter')
+
+        # Make sure 1e6 doesn't get a prefix in it
```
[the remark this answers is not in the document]
MichaelClerx: Good idea!

tests/test_units.py
MichaelClerx: Good point!

tests/test_units.py
MichaelClerx: Not sure I understand the need for the last 3?

MichaelClerx:
Ready for re-review once the tests have passed!

tests/test_units.py
jonc125: It's basically making sure the regex replacement doesn't do anything silly.

tests/test_units.py
```diff
@@ -20,14 +21,19 @@ def test_add_unit(self):
         # Add ordinary unit
         unitstore = UnitStore()
         assert not unitstore.is_defined('u1')
-        unitstore.add_unit('u1', 'second * 2')
+        u1 = unitstore.add_unit('u1', 'second * 2')
         assert unitstore.is_defined('u1')
-        assert unitstore.get_unit('u1') == unitstore.get_unit('second') * 2
+        assert u1 == unitstore.get_unit('second') * 2
```
jonc125: Maybe also `assert u1 is unitstore.get_unit('u1')`?

tests/test_units.py
```diff
@@ -85,18 +98,72 @@ def test_is_equivalent(self):
...
+        assert store.get_conversion_factor(s, ms) == 0.001
+        assert store.get_conversion_factor(ms, s) == 1000
+
+    def test_prefixing(self):
+        """Tests edge cases for the internal prefixing."""
+
+        store = UnitStore()
+        x = store.add_unit('x', 'meter')
+
+        # Use str(x) as a new unit name: this will have any prefix/postfix the unit store adds so could leave to
```
jonc125: leave -> lead

tests/test_units.py
MichaelClerx: Weirdly, it has to be `==`. Repeated calls to `get_unit()` don't return the same object.

PR #225, opened 2020-02-12
Assume real for dummies

MauriceHendrix:
## Description

For NumberDummy and VariableDummy assume values are real

## Motivation and Context

If we don't make this assumption sympy starts brining in imaginaries in certain derivatives. Since cellml variables are real we can safely assume this

## Types of changes

## Checklist:

## Testing


MauriceHendrix:
> This is fine, but might be nice to add a test that it achieves something? Can you check `var.is_real` or similar?

was just doing that. I thought I might try to do it through the web interface :)

Issue #228, opened 2020-02-17
Maintain a cmeta_id to symbol mapping


jonc125:
Looking up a symbol by ontology term relies on lookup by cmeta_id, and we do a lot of these in Web Lab logic (and several in Chaste code gen too). But lookup by cmeta_id currently requires iterating over all symbols. Looking up in a dict would be better. We just need to ensure the map can't get out of sync too easily - at present updating cmeta ids is just an operation on the `VariableDummy` objects so the `Model` doesn't know about it. To fix this we'll need 2 new methods on `Model`:
* [x] `transfer_cmeta_id(self, from, to)` moves an ID from one variable to another. Used in units conversion code. Just needs to move the `cmeta_id` from `from` to `to` and update that entry in the map.
* [x] `add_cmeta_id(self, var)` adds an ID to a variable that doesn't have one. It can be implemented as `var.cmeta_id = self.get_unique_cmeta_id(var.name.replace('$', '__'))` (and add to the map)

MichaelClerx:
To make it safe, we have to stop users from inadvertently changing a cmeta id.
Three options:

1. Make a `cmeta_id()` method. My preferred option as it's clear to everyone that it exists and what it does.
2. Use the @property decorator to make `cmeta_id` a read-only property. Document this in the class's main docstring.
3. Use @property but go all out and let users set it and then automatically update the model (requires to have a link to the model in each VariableDummy)

MichaelClerx:
We've already got a `rdf_identity` property in there, which is a plus for option 2 I guess?

jonc125:
With the `rdf_identity` property there's probably very little need for external code to be reading the `cmeta_id`, so could make it an internal property that only the `Model` sets?

MichaelClerx:
ok!

MichaelClerx:
Then again, that's a bit weird, we've got all these methods to transfer cmeta ids but the user isn't allowed see them?

MichaelClerx:
Hmmmm, there's also a lot of RDF manipulation going on, but the documentation simply says "further meta data" is stored in the RDF graph. How do I find out what meta data and whether or not I need to do anything with this when updating cmeta ids?

jonc125:
The point of transferring IDs is that the metadata then doesn't need updating. You're just changing the variable it points at.

We could have a readonly public property to let people read the ID, just not (easily) set it. But it's really the RDF term that Web Lab use cases need; the ID is just internal detail.

[...]

PR #229, opened 2020-02-19
Store mapping from cmeta ids to variable dummies

MichaelClerx:
## Description
See #228

## Types of changes
- Breaking change (fix or feature that would cause existing functionality to change)

## Testing
- Testing is done automatically and codecov shows test coverage

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -39,6 +39,9 @@ class Model(object):
 
     Units are handled using the ``units`` property of a model, which is an instance of ``cellmlmanip.units.UnitStore``.
 
+    RDF meta data can be attached to either the model itself or to model variables, via the CellML ``cmeta:id``
+    attribute. Cmeta ids set on any other parts of CellML models are ignored.
+
     Cellmlmanip does not support algebraic models: the left-hand side every equation in the model must be a variable or
     a derivative.
 
@@ -94,10 +97,13 @@ class Model(object):
             return
 
         # Create new cmeta id
-        variable._cmeta_id = self.get_unique_cmeta_id(str(variable))
+        cmeta_id = str(variable)
+        while self.has_cmeta_id(cmeta_id):
+            cmeta_id += '_'
 
-        # Store mapping from id to dummy
-        self._cmeta_id_to_symbol[variable._cmeta_id] = variable
+        # Add to variable and store in mapping
+        variable._cmeta_id = cmeta_id
+        self._cmeta_id_to_symbol[cmeta_id] = variable
 
     def add_equation(self, equation, check_duplicates=True):
         """
@@ -175,11 +181,9 @@ class Model(object):
         if name in self._name_to_symbol:
             raise ValueError('Variable %s already exists.' % name)
 
-        # Check uniqueness of cmeta id (against RDF graph)
-        # if cmeta_id is not None:
-        #    cmeta_id = str(cmeta_id)
-        #    if next(self.rdf[create_rdf_node('#' + cmeta_id)], None) is not None:
-        #        raise ValueError('The cmeta_id "%s" is already in use.' % cmeta_id)
+        # Check uniqueness of cmeta id
+        if cmeta_id is not None and self.has_cmeta_id(cmeta_id):
+            raise ValueError('The cmeta_id "%s" is already in use.' % cmeta_id)
 
         # Check units
         if not isinstance(units, self.units.Unit):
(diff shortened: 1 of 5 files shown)


MichaelClerx:
Ready for review

cellmlmanip/model.py
```diff
@@ -1048,4 +1106,9 @@ def __str__(self):
     @property
     def rdf_identity(self):
         """The RDF identity for this variable, or ``None`` if no cmeta id."""
-        return create_rdf_node('#' + self.cmeta_id) if self.cmeta_id else None
+        if self._cmeta_id is None:
+            return None
+        if self._rdf_identity is None:
+            self._rdf_identity = create_rdf_node('#' + self._cmeta_id)
+        return self._rdf_identity
```
jonc125: Because this is now cached, if a user has accessed `rdf_identity` then transfers the cmeta_id away, *both* variables will end up with the same `rdf_identity` cached.

I think it'd be safer to add a `_set_cmeta_id()` method on `VariableDummy` that sets both `self._cmeta_id` and `self.rdf_identity`, and remove this `@property` method.

MichaelClerx: :flushed: woops!

tests/test_rdf.py
```diff
@@ -211,6 +211,48 @@ def test_add_rdf(local_model):
...
+    rid = v.rdf_identity
+    assert rid is not None
+    model.add_cmeta_id(v)
+    assert v.rdf_identity == rid
+
+    # Test on variable with a name that's already in use as a cmeta id
+    assert model.has_cmeta_id('time')
+    v = model.add_variable('time', 'second')
+    model.add_cmeta_id(v)
+
```
jonc125: This bit doesn't actually test anything!

tests/test_model_functions.py
```diff
@@ -547,6 +554,19 @@ def test_bad_units(self, bad_units_model):
             # cellml file states b (per_ms) = power(a (ms), 1 (second))
             bad_units_model.units.evaluate_units(equation[1].rhs)
 
+    ###################################################################
+    # rdf related methods
+
+    def test_has_cmeta_id(self, basic_model):
```
jonc125: Is there a reason this is here rather than test_rdf.py?

tests/test_rdf.py
```diff
@@ -211,6 +211,48 @@ def test_add_rdf(local_model):
...
+    model = shared.load_model('test_simple_odes')
+
+    # Legal move: from variable with a cmeta id to one without
+    v1 = model.get_symbol_by_ontology_term((shared.OXMETA, 'time'))
+    v2 = model.get_symbol_by_name('circle_x$x')
+    assert v1.rdf_identity is not None
+    assert v2.rdf_identity is None
+    model.transfer_cmeta_id(v1, v2)
+    assert v1.rdf_identity is None
```
jonc125: Interesting that this passed given my earlier comment! I guess because the property checks `_cmeta_id` first.

tests/test_rdf.py
```diff
@@ -211,6 +211,48 @@ def test_add_rdf(local_model):
...
+    assert v2.rdf_identity is not None
+
+    # Illegal move: to a variable with a cmeta id
+    v3 = model.get_symbol_by_ontology_term((shared.OXMETA, 'cytosolic_calcium_concentration'))
+    assert v2.rdf_identity is not None
+    assert v3.rdf_identity is not None
+    with pytest.raises(ValueError):
+        model.transfer_cmeta_id(v2, v3)
+
```
jonc125: Should we also check that you can't transfer away from a variable without a cmeta id?

tests/test_rdf.py
MichaelClerx: It tests that you can call add_cmeta_id on a variable that already has a cmeta id without getting an exception

tests/test_rdf.py
MichaelClerx: Why shouldn't this pass?

tests/test_rdf.py
jonc125: Indeed, but it doesn't check that the rdf_identity is the same afterwards, and probably should?

tests/test_rdf.py
jonc125: It should! I'd just assumed because self._rdf_identity is cached that it wouldn't, but rdf_identity has an extra check before accessing the cache which is why it does currently. That check won't be needed now.

tests/test_rdf.py
MichaelClerx: The problem is it's called on a variable without a cmeta id, but one for which the normal strategy of using the variable name fails.

This strategy isn't really part of the public API though! How the clash is resolved definitely isn't, so shouldn't be checking that here?

MichaelClerx: Sorry my comment above was wrong!

MichaelClerx:
Windows tests failing due to https://github.com/ModellingWebLab/cellmlmanip/issues/233

tests/test_rdf.py
jonc125: Ah, good point. So don't check the identity is unchanged, but perhaps do check that it's not None and not the same as for the existing time variable?

MichaelClerx: Good idea!

PR #237, opened 2020-02-24
Renamed _symbol_ methods to _variable_ methods.

MichaelClerx:
## Description
Renamed _symbol_ methods to _variable_ methods.

## Motivation and Context
#230 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have updated all documentation necessary.
- [x] I have checked spelling in (new) comments.

## Testing
- [X] Testing is done automatically and codecov shows test coverage
- [ ] This cannot be tested automatically

PR #239, opened 2020-02-25
Added Maurice' fix and test for transfer cmeta id if the original variable doesn't have a cmeta id.

MichaelClerx:
Copied @MauriceHendrix 's changes from #237


MichaelClerx:
This looks fine to me.

MichaelClerx:
Thanks for the bugfix, @MauriceHendrix

PR #244, opened 2020-02-25
243 transform units bug

### LINKED ISSUE #243: Transform constants doesn't add units
```
<!-- CellML 1.0, 3.4.6.4: Siblings are connected via the public interface -->
<!-- B.a to C.a, both children of A -->
<model name="map_variables_nested_sibling_connection"
  <component name="A" />
  <component name="B">
    <variable name="a" units="newton" public_interface="out" initial_value="1" />
  </component>
  <component name="C">
    <variable name="a" units="newton" public_interface="in" />
  </component>
  <group>
    <relationship_ref relationship="encapsulation" />
    <component_ref component="A">
      <component_ref component="B" />
      <component_ref component="C" />
    </component_ref>
  </group>
  <connection>
    <map_components component_1="B" component_2="C" />
    <map_variables variable_1="a" variable_2="a" />
  </connection>
</model>
```

gives the error

```
Units newton 1.0 kilogram * meter / second ** 2 != dimensionless 1 dimensionless
```

Not sure what's happening here. Accidental unit conversion?

MichaelClerx:
Fixes #243 

## Types of changes

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -248,7 +248,9 @@ class Model(object):
 
         # If the source variable has not been assigned a value, we can't make this connection
         if not source.assigned_to:
-            logger.info('The source variable has not been assigned a value yet', target.name, source.name)
+            logger.debug(
+                'Cannot connect %s to %s at this time: The source variable has not been assigned a value '.format(
+                    target.name, source.name))
             return False
 
         # If target is already assigned this is an error
@@ -556,9 +558,12 @@ class Model(object):
                 # Get the free symbol and update the variable information
                 free_symbol = lhs.variables[0]
                 free_symbol.type = 'free'
-            elif equation.rhs.is_number:
+            elif isinstance(equation.rhs, NumberDummy):
                 lhs.type = 'parameter'
+                # logger.error('Parameter: %s = %s', equation.lhs, equation.rhs)
             else:
+                # logger.error('Computed: %s = %s', equation.lhs, equation.rhs)
+
                 lhs.type = 'computed'
 
         # Sanity check: none of the lhs have the same hash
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Weirdly, I now see a new error from `TestModelFunctions.test_get_derived_quantities`

```
Message: 'The source variable has not been assigned a value yet'
Arguments: ('circle_sibling$y', 'circle_parent$y')
--------------------------------- Captured log setup ----------------------------------
INFO     cellmlmanip.model:model.py:293 Connection req. unit conversion: Eq(_time_units_conversion1$time, _0.001*_environment$time)
INFO     cellmlmanip.model:model.py:293 Connection req. unit conversion: Eq(_time_units_conversion2$time, _1000.0*_environment$time)
```

jonc125:
> Weirdly, I now see a new error from `TestModelFunctions.test_get_derived_quantities`

I think that's because you got rid of some %s tokens in the logging in an earlier PR #237 !
`logger.info('The source variable has not been assigned a value yet', target.name, source.name)`

MichaelClerx:
Oh dear!

MichaelClerx:
> I think that's because you got rid of some %s tokens in the logging in an earlier PR #237 !
logger.info('The source variable has not been assigned a value yet', target.name, source.name)

I think I must have accidentally hit the keyboard shortcut for deleting a line :flushed:

jonc125:
Though I think the failing test must be because the change to `transform_constants` means the LHS no longer gets marked as a parameter in the graph, presumably because `equation.rhs.is_number` is False. Do we need to tweak `NumberDummy` so it claims to be a number to Sympy?

MichaelClerx:
We're not meant to have any numbers at that point though, so perhaps it should be an `isinstance(NumberDummy)` instead?

MichaelClerx:
(Also worried having sympy treat our number dummies as numbers would bring back problems we're avoiding by having them _not_ being numbers)

MichaelClerx:
Seems to me the current test is wrong!

It shows `single_ode_rhs_computed_var$a = -1.0` should be marked as a derived quantity.

Paging @MauriceHendrix just in case that was intended behaviour

PR #251, opened 2020-02-27
added class level method to change how mathml operators are handled

MauriceHendrix:
## Description

Added class level method to be able to change how the transpiler handles mathml operators

## Motivation and Context

I'm currently directly messing with the internals from chaste codegen so I can control the output better, but that is a bit hacky

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/transpiler.py
+++ b/cellmlmanip/transpiler.py
@@ -54,16 +54,17 @@ class Transpiler(object):
         }
 
         # Add tags that can be handled by simple_operator_handler
-        for tag_name in SIMPLE_MATHML_TO_SYMPY_NAMES:
+        for tag_name in SIMPLE_MATHML_TO_SYMPY_CLASSES:
             self.handlers[tag_name] = self._simple_operator_handler
 
+    @staticmethod
     def set_mathml_handler(mathml_operator, operator_class):
         """Change how the transpiler handles a given mathml_operator.
 
         :param mathml_operator: The name of a MathML operator e.g. 'exp', 'true' etc.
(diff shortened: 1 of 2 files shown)


cellmlmanip/transpiler.py
```diff
@@ -58,6 +57,14 @@ def __init__(self, symbol_generator=None, number_generator=None):
         for tag_name in SIMPLE_MATHML_TO_SYMPY_NAMES:
             self.handlers[tag_name] = self._simple_operator_handler
 
+    def set_mathml_handler(mathml_operator, operator_class):
```
jonc125: I think you want an `@staticmethod` decorator for this?

cellmlmanip/transpiler.py
```diff
@@ -58,6 +57,14 @@ def __init__(self, symbol_generator=None, number_generator=None):
...
+        """Change how the transpiler handles a given mathml_operator.
+
+        :param mathml_operator: The name of a MathML operator e.g. 'exp', 'true' etc.
+        :param operator_class: A class that can handle the given operator e.g. sympy.exp, sympy.true etc.
+        """
+        SIMPLE_MATHML_TO_SYMPY_NAMES[mathml_operator] = operator_class
```
jonc125: Given the values are no longer names, this dict should be renamed e.g. to `SIMPLE_MATHML_TO_SYMPY_FUNCTIONS`

jonc125: Though they're not all functions either...

MauriceHendrix: technically they are classes actually

Issue #252, opened 2020-02-28
using replace/xreplace instead of subs


MauriceHendrix:
I just wondered if there were any objections to changing model.py so that it uses xreplace and replace instead of subs? Replace is considerably faster as it doesn't try to understand the maths (e.g. you can do .replace(x, 0.0) but if you wanted to update an expression with something more comples as query you'd need subs. However it looks we're always just replacing a symbol (or derivative).

To see what I mean in code see #253

PR #253, opened 2020-02-28
replaces subs with replace/xreplace

MauriceHendrix:
## Description

Replaces subs with replace/xreplace

## Motivation and Context

This leads to somewhat better performance especially in some parts of the code generation
see #252

## Types of changes

## Checklist:

- [ ] I have updated all documentation necessary.
n/a

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -630,7 +630,7 @@ class Model(object):
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
                 # Update rhs
-                rhs = equation.rhs.xreplace(subs_dict)
+                rhs = equation.rhs.subs(subs_dict)
 
                 # Check if simplification removed dependencies on other variables, and if so remove the corresponding
                 # edges.
@@ -887,7 +887,7 @@ class Model(object):
             for argument in equation.rhs.atoms(sympy.Derivative):
                 if new_derivative['expression'] == argument:
                     # add new equation
-                    new_eqn = equation.replace(new_derivative['expression'], new_derivative['variable'])
+                    new_eqn = equation.xreplace({new_derivative['expression']: new_derivative['variable']})
                     self.remove_equation(equation)
                     self.add_equation(new_eqn)
                     break
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -631,7 +631,7 @@ class Model(object):
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
                 # Update rhs
-                rhs = equation.rhs.subs(subs_dict)
+                rhs = equation.rhs.xreplace(subs_dict)
 
                 # Check if simplification removed dependencies on other variables, and if so remove the corresponding
                 # edges.
@@ -888,7 +888,7 @@ class Model(object):
             for argument in equation.rhs.atoms(sympy.Derivative):
                 if new_derivative['expression'] == argument:
                     # add new equation
-                    new_eqn = equation.subs(new_derivative['expression'], new_derivative['variable'])
+                    new_eqn = equation.replace(new_derivative['expression'], new_derivative['variable'])
                     self.remove_equation(equation)
                     self.add_equation(new_eqn)
                     break
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -630,7 +630,7 @@ class Model(object):
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
                 # Update rhs
-                rhs = equation.rhs.subs(subs_dict)
+                rhs = equation.rhs.xreplace(subs_dict)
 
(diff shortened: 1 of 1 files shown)


cellmlmanip/model.py
```diff
@@ -888,7 +888,7 @@ def _replace_derivatives(self, new_derivative):
             for argument in equation.rhs.atoms(sympy.Derivative):
                 if new_derivative['expression'] == argument:
                     # add new equation
-                    new_eqn = equation.subs(new_derivative['expression'], new_derivative['variable'])
```
jonc125: We're not using "wildcard-like matching, parsing of match, and conditional replacements" here, so why not use `xreplace`?

cellmlmanip/model.py
```diff
@@ -631,7 +631,7 @@ def graph_with_sympy_numbers(self):
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
                 # Update rhs
-                rhs = equation.rhs.subs(subs_dict)
```
jonc125: This does change behaviour, since `subs` will simplify whereas `xreplace` will not. So if e.g. we get `0 * expr` after substitution then with `subs` we could remove dependencies on other variables, hence the need for the following lines removing edges.

Possibly that's ok, but it definitely needs discussion. And we should also check whether the following lines can indeed be removed with this change.

MauriceHendrix: Actually xreplace will also simplify or at least it did in my test cases, despite the docs not mentioning it.

cellmlmanip/model.py
MauriceHendrix: xreplace takes a dictionary, and replace takes 2 arguments. As far as I understand it the no wildcard equivalent for `.subs(old, new`) is `.replace(old, new)` and the no-wildcard equivalent for `.subs({old1, new1, ..., oldn: newn})` is `.xreplace({old1, new1, ..., oldn: newn})` is

cellmlmanip/model.py
jonc125: OK, maybe just more limited canonicalisation? Not sure! I'm surprised that the apparently more complex `replace` would be faster.

Could also consider the `simultaneous=True` argument to `subs`.

jonc125: yeah, looking at the code for `xreplace` it is just recursively doing `self.func(*args)` after calling xreplace on its arguments, so the only simplification is that done by construction.

jonc125: Looking at the code for subs, it will call `sympify` explicitly on both old & new items, which will be a performance hit. It also does a lot more complex recursion down the tree than does `xreplace`.

`replace` will also `_sympify` its arguments and do more complex recursion.

cellmlmanip/model.py
jonc125: Neither `subs` nor `xreplace` uses wildcards; only `replace` does that!

You can replace `subs(old, new)` with `xreplace({old: new})` if you don't need to worry about bound variables.

Where it might go wrong is if the 'old' was a state variable or time, since `xreplace` would change derivatives which we might not want. Though sometimes this might be intended!

Another case `xreplace` can't do is replacing e.g. `x*y` within `x*y*z`, because it only finds complete nodes in the expression tree that match, not partial nodes.

Here though we're replacing a derivative itself with a variable, so `xreplace` should be fine.

MichaelClerx:
I don't really mind what happens here :D

cellmlmanip/model.py
MauriceHendrix: Ok I see

cellmlmanip/model.py
MauriceHendrix: Ok I reverted to subs here

cellmlmanip/model.py
```diff
@@ -630,7 +630,7 @@ def graph_with_sympy_numbers(self):
             # And replace the equation with one with the rhs subbed with sympy.Number objects
             if subs_dict:
                 # Update rhs
-                rhs = equation.rhs.xreplace(subs_dict)
+                rhs = equation.rhs.subs(subs_dict)
```
jonc125: I thought `xreplace` was OK here?

jonc125: Or is `subs` somehow faster than `xreplace`?

jonc125: Would love some numbers to see what the difference in speed between the various options is :)

MauriceHendrix: @jonc125 right so here is what I did:
put the derivative eqs & symbols for the bondarenko_model_2004_apex model in a python script and perform a partial_eval style replacement (everything substituted into the derivative equations). 10 times for each method on my pc. The following is in time.process_time()
xreplace took:12.015625
subs took:126.5
replace (in a loop over subs_dict) took:29.234375
xreplace (in a loop over subs_dict) took:28.734375

MauriceHendrix: (changed it to xreplace)

PR #256, opened 2020-03-06
Updated docstring for convert_variable.

MichaelClerx:
Made some changes to a docstring that hopefully make it more clear?

EDIT: And some code changes

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -874,7 +874,7 @@ class Model(object):
         assert isinstance(units, self.units.Unit)  # Units must be in the right registry
         assert isinstance(direction, DataDirectionFlow)
 
-        # Compute conversion factor for old units to new
+        # Compute conversion factor for old units to new;
         # throws DimensionalityError if unit conversion is not possible
         cf = self.units.get_conversion_factor(from_unit=original_variable.units, to_unit=units)
         if cf == 1:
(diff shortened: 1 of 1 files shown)


MichaelClerx:
I'm confused by the syntax of the pseudo code as well

    var{time} time: ms {pub: in};

Why does it say time twice?
And why `sv11` here?

    var{sv11} sv1: mV {init: 2};

MichaelClerx:
And why do we have semicolons :D :D :D

MichaelClerx:
Also the `time` case is not something I think users would consider straight away, so wouldn't have it as the first example. Maybe would start with an OUTPUT, then a (non-time) INPUT, and only then time?

jonc125:
The syntax is a bit arbitrary! I think the `{time}` part gives the annotation, so it's showing that that moves as needed. The examples were taken from our main test model.

Re-ordering the examples and simplifying the syntax is fine by me.

MichaelClerx:
I've tried tidying up the code a bit too. @skeating could you maybe have a go at cleaning this up as well, now that the tests are all there and it's clearer what the code should do?

I tried reading it all but it's quite difficult because
- some things are quite split up: state values are overdefined for a while, and this situation is created in one method, then fixed in another, and it requires a special `check_duplicates=False` trick in `add_variable` which doesn't seem to be used everywhere else
- the bit where a dict gets passed around were tricky to read, in part because of the entries 'variable' and 'expression' when there are 3 variables and several expressions involved
- Does it really need to be split over so many methods? I wouldn't mind if `_create_new_deriv_variable_and_equation` wasn't there, for example, even if that caused a bit of duplicate code. It'd maybe make more sense to split off time variable and state variable conversion into separate methods, so at the end of `convert_variable` have an `if free: method x; elif state: method y`?
- the bit where refs to derivatives in the model RHS are handled could probably be optimised a bit for the time case? So use the current code for state vars, but for time vars create a single dict with all derivs that need replacing, pass it into xreplace, and replace the equation if the result is different from the original eq?

Just some thoughts! Probably someone else has a better idea of how to tidy this a bit that's why I'm asking instead of attempting it myself

skeating:
Discussed this with @MichaelClerx and I will take it over and try and clarify documentation.

jonc125:
@MichaelClerx I'm happy that this looks better now. As you authored the PR I can't ask you for a formal review though! So I suggest you merge if you like it :)

cellmlmanip/model.py
```diff
@@ -794,52 +794,71 @@ def variables(self):
...
+        If the direction is ``INPUT`` then any initial value will be moved to the new variable (and converted
+        appropriately).
 
-        For example::
+        For example, a model::
 
-            Original model
-                var{time} time: ms {pub: in};
-                var{sv11} sv1: mV {init: 2};
+            var time :: ms {meta: time}
```
MichaelClerx, line 814: Shall we just write `cmeta_id` for `meta` ?

jonc125: Could do!

MichaelClerx:
OK, looks better!

PR #274, opened 2020-04-27
added model.get_display_name

MauriceHendrix:
## Description

Added model.get_display_name for alpahabetical ordering or a list of variables, first by ontology term, then cmeta id and finally name

## Motivation and Context

In various places in fc and codegen we need a consistant reliable ordering, that's also consitant with the way pycml used to do it.

## Types of changes

## Checklist:

## Testing


cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
                               and node.get('variable_type', '') not in ('state', 'free', 'parameter')]
         return sorted(derived_quantities, key=lambda var: var.order_added)
 
+    def get_display_name(self, var, META='https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#'):
```
jonc125, line 390: Method arguments shouldn't be all-caps. That's for constants.

I'd called this `preferred_ontology` or similar.

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+        """Return a display name for the given variable.
+
+        Looks for META ontology annotation tag first, then cmeta:id if present, or the name attribute if not.
+        If there is an interface component, strip the name of it out of the display name.
+        :param var: the variable for which to get the display name.
+        :param META: the namespace prefix for the ontology used, (by default the oxford chaste metadata ontology).
+
+        :return: ontology annotation if the var has it, otherwise cmeta_id, if it has it otherwise variable name.
```
jonc125, line 398: Perhaps "the local name from the preferred ontology (if such an annotation exists), or the cmeta_id (if present) or the variable's name"

tests/cellml_files/test_display_name.cellml
```diff
@@ -0,0 +1,682 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jonc125, line 1: We can probably reuse one of the existing test models rather than adding a new one?

cellmlmanip/model.py
MichaelClerx, line 390: Should `META` be called `ontology` ?

tests/test_model_functions.py
```diff
@@ -84,6 +84,24 @@ def test_get_derived_quantities(self, basic_model, simple_ode_model):
             '_deriv_on_rhs2b$sv1_rate]'
         )
 
+    def test_get_display_name(self):
```
jonc125, line 87: I'd like lower-level tests here. So check not all states, but individual variables covering the cases:
* Has an oxmeta-annotation
* Has an annotation in another ontology
* Has no annotation but does have a cmeta:id
* Has no cmeta:id

cellmlmanip/model.py
MichaelClerx: I'd prefer it if the method didn't select a particular ontology at all if you didn't specify this.

So 
- `get_display_name(self, var, ontology=None)`
- To get only terms from oxmeta, call with `get_display_name(self, var, 'https://....')`
- To get any old annotation, call with `get_display_name(self, var)`

tests/cellml_files/test_display_name.cellml
MauriceHendrix: found one

tests/test_model_functions.py
MauriceHendrix: Ok I checked 4 individual variables for these for cases in test_simple_odes.cellml instead

cellmlmanip/model.py
MichaelClerx: @MauriceHendrix see previous comment above

MauriceHendrix: hhmm ok fine although my calls are going to be longer now

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
                               and node.get('variable_type', '') not in ('state', 'free', 'parameter')]
         return sorted(derived_quantities, key=lambda var: var.order_added)
 
+    def get_display_name(self, var, ontology=None):
+        """Return a display name for the given variable.
+
+        Looks for annotation in the ontology first, then cmeta:id if present, or the name attribute if not
```
jonc125, line 393:
```suggestion
        Looks for annotation in the ontology first (if given), then cmeta:id if present, or the name attribute if not
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+        If there is an interface component, strip the name of it out of the display name
```
jonc125, line 394: Cut this line since we don't do this.
```suggestion
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+        :param var: the variable for which to get the display name.
+        :param ontology: the namespace prefix for the ontology used.
```
jonc125, line 396:
```suggestion
        :param ontology: the namespace prefix for the ontology used (defaults to not considering annotations).
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+
+        :return: the local name from the ontology (if such an annotation exists)
+                 or the cmeta_id (if present) or the variable's name"""
```
jonc125, line 399:
```suggestion
        :return: the local name from the ontology (if such an annotation exists)
                 or the cmeta_id (if present) or the variable's name
        """
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+        if self.has_ontology_annotation(var, ontology):
```
jonc125, line 400:
```suggestion
        if ontology is not None and self.has_ontology_annotation(var, ontology):
```

tests/test_model_functions.py
```diff
@@ -84,6 +84,51 @@ def test_get_derived_quantities(self, basic_model, simple_ode_model):
...
 
+    def test_get_display_name(self, simple_ode_model):
+        """ Tests Model.get_display_name(var). """
+        # For getting a display name there are 4 possibilities:
+        # - Has an oxmeta-annotation
+        # - Has an annotation in another ontology
+        # - Has no annotation but does have a cmeta:id
+        # - Has no cmeta:id
+
+        # Check our 4 example variables exist
```
jonc125, line 95: Rather than this long-winded approach, can't you just do e.g.
```python
    var_oxmeta = simple_ode_model.get_variable_by_name('single_independent_ode$sv1')
    assert var.cmeta_id == 'sv11'
    assert simple_ode_model.get_display_name(var, ontology=OXMETA) == 'sodium_reversal_potential'
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,20 @@ def get_derived_quantities(self):
...
+        """Return a display name for the given variable.
+
+        Looks for annotation in the ontology first (if given), then cmeta:id if present, or the name attribute if not
+        :param var: the variable for which to get the display name.
+        :param ontology: the namespace prefix for the ontology used (defaults to not considering annotations).
+
+        :return: the local name from the ontology (if such an annotation exists)
+                 or the cmeta_id (if present) or the variable's name
+        """
+        if ontology is not None and self.has_ontology_annotation(var, ontology):
```
MichaelClerx, line 400: Why the `ontology is not None` here? I want it to return from _any_ ontology if I don't specify a specific one?

cellmlmanip/model.py
MichaelClerx: Wait @jonc125 is this what we want? Would we not prefer it to return (1) a specific ontology if given, else an annotation from _any_ ontology (2) if 1 fails, then a cmeta:id, (3) a name attribute

cellmlmanip/model.py
MichaelClerx: @jonc125 is this different from what you had in mind?

cellmlmanip/model.py
MauriceHendrix: Hhm actually it should look also for non, meaning "any old ontology"

cellmlmanip/model.py
MauriceHendrix: I just applied Jonathan's comment without thinking. I reverted it now!

MichaelClerx:
Cool. I like this version. But let's see what @jonc125 says tomorrow morning :D

cellmlmanip/model.py
jonc125: Ah, sorry, I thought the request before was to do this only if an ontology was specified! Just me misunderstanding the request.

cellmlmanip/model.py
```diff
@@ -387,6 +387,22 @@ def get_derived_quantities(self):
                               and node.get('variable_type', '') not in ('state', 'free', 'parameter')]
         return sorted(derived_quantities, key=lambda var: var.order_added)
 
+    def get_display_name(self, var, ontology=None):
+        """Return a display name for the given variable.
+
+        Looks for annotation in the ontology first (if given), then cmeta:id if present, or the name attribute if not
```
jonc125, line 393:
```suggestion
        Looks for an annotation in the ontology first (or the local name from any annotation if no ontology is
        specified), then cmeta:id if present, or the variable's name attribute if not.
        
        Dollar symbols in the name are replaced by a double underscore.
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,22 @@ def get_derived_quantities(self):
...
+        :param var: the variable for which to get the display name.
+        :param ontology: the namespace prefix for the ontology used (defaults to not considering annotations).
+
+        :return: the local name from the ontology (if such an annotation exists)
+                 or the cmeta_id (if present) or the variable's name
```
jonc125, line 398:
```suggestion
        :return: the display name for the variable according to the algorithm above
```

cellmlmanip/model.py
```diff
@@ -387,6 +387,22 @@ def get_derived_quantities(self):
                               and node.get('variable_type', '') not in ('state', 'free', 'parameter')]
         return sorted(derived_quantities, key=lambda var: var.order_added)
 
+    def get_display_name(self, var, ontology=None):
+        """Return a display name for the given variable.
+
+        Looks for annotation in the ontology first (if given), then cmeta:id if present, or the name attribute if not
+        :param var: the variable for which to get the display name.
+        :param ontology: the namespace prefix for the ontology used (defaults to not considering annotations).
```
jonc125, line 395:
```suggestion
        :param ontology: the base URL of an ontology if only annotations within that ontology should be considered
```

tests/test_model_functions.py
```diff
@@ -84,6 +84,33 @@ def test_get_derived_quantities(self, basic_model, simple_ode_model):
...
+        var = simple_ode_model.get_variable_by_name('single_ode_rhs_const_var$a')
+        assert var.cmeta_id == 'a1'
+        assert simple_ode_model.get_display_name(var, ontology=OXMETA) == var.cmeta_id
+        assert simple_ode_model.get_display_name(var) == 'parameter_a1'
+        assert simple_ode_model.get_display_name(var, ontology='urn:test-ns#') == 'parameter_a1'
+
+        # Has no annotation but does have a cmeta:id
+        var = simple_ode_model.get_variable_by_name('single_ode_rhs_computed_var$a')
+        assert var.cmeta_id == 'a2'
+        assert simple_ode_model.get_display_name(var, ontology=OXMETA) == var.cmeta_id
```
jonc125, line 107:
```suggestion
        assert simple_ode_model.get_display_name(var, ontology=OXMETA) == var.cmeta_id
        assert simple_ode_model.get_display_name(var) == var.cmeta_id
```

tests/test_model_functions.py
```diff
@@ -84,6 +84,33 @@ def test_get_derived_quantities(self, basic_model, simple_ode_model):
...
+
+        # Has no cmeta:id
+        var = simple_ode_model.get_variable_by_name('single_ode_rhs_const_var$time')
+        assert var.cmeta_id is None
+        assert simple_ode_model.get_display_name(var, ontology=OXMETA) == var.name.replace('$', '__')
```
jonc125, line 112:
```suggestion
        assert simple_ode_model.get_display_name(var) == var.name.replace('$', '__')
```

Issue #120, opened 2019-11-15
Split CellML parsing & model API code more thoroughly?


MichaelClerx:
The current Model class handles part of the CellML parsing, via `connect_variables()` and interface objects in the MetaDummy class. Moving this into the parser would make the model API easier to use / understand, but also pave the way for CellML 2.0 support

See also e.g. #81

> Can we do this ignoring the directionality on interfaces? (Makes future CellML 2 support easier.)

Thoughts?

jonc125:
My approach would be to iteratively improve the API as needed to implement protocol features, without trying to fix it too much up-front; I don't think we understand the requirements well enough yet. Particularly the directionality aspect has more impact on the dependency graph generation than on the parsing I think.

MichaelClerx:
But you collapse the two variables into one, in CellML 2, so there's only a single entity and no scope for directionality

MichaelClerx:
Anyway this issue arose from

> iteratively improv[ing] the API as needed to implement protocol features

because `add_variables` looks like this:

    def add_variable(self, *, name, units, initial_value=None,
                     public_interface=None, private_interface=None, **kwargs):
   
and now I'm trying to understand if I need to do anything with those public and private interfaces if I want to add a variable or not

jonc125:
Which we do already where there is no unit conversion, but if there is, you need to add an equation. And which variable do you make the LHS? Currently we assume all equations are assignments. But in general in CellML (even in v1! Just not in most cardiac models :grin:) equations are just equality declarations, and in code generation information might flow either way; there might even be systems of simultaneous equations to solve.

MichaelClerx:
Anyway this issue arose from

> iteratively improv[ing] the API as needed to implement protocol features

because `add_variable` looks like this:

    def add_variable(self, *, name, units, initial_value=None,
                     public_interface=None, private_interface=None, **kwargs):
   
and now I'm trying to understand if I need to do anything with those public and private interfaces if I want to add a variable or not

jonc125:
> and now I'm trying to understand if I need to do anything with those public and private interfaces if I want to add a variable or not

I'd hope not after the initial parse...

tamuri:
> I'd hope not after the initial parse...

Yeah, those are only used for resolving the connections. If you're adding a variable after loading the model, you don't have to give anything (unless you're intending to run connect variables again)

skeating:
[...]

PR #276, opened 2020-04-30
Tidying

### LINKED ISSUE #161: Rename/move some Model functions
1. Model. add_number()
This function merely creates a NumberDummy object. It is completely independent of the Model and adds nothing to anything. I think the name could be misleading.

I would prefer: create_number_symbol 

2. Model.check_left_right_units_equal(sympy.Equality)
This functions should be in the one of the unit classes (UnitCalculator); although that would involve passing the model/model units as an argument - so I could let it stay; but it does feel wrong. 

3. Model.find_symbols_and_derivatives(expression)
This creates a set of symbols from the list of expressions passed to it. It does not find symbols/derivatives that are in the model. Again possible misleading.  Also not sure this belongs on Model class (same as add_number) they are helper functions really.

jonc125:
## Description

Doing various bits of tidying/refactoring in preparation for initial release.

## Motivation and Context

Does some of #161 and #120. Probably others before I'm done.

## Types of changes
- [x] Refactoring

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -409,10 +409,9 @@ class Model(object):
 
     def get_free_variable(self):
         """Returns the free variable in this model (if any)."""
-        # TODO: Use self._ode_definition_map instead?
-        for v, node in self.graph.nodes.items():
-            if node.get('variable_type', VariableType.UNKNOWN) == VariableType.FREE:
-                return v
+        for ode in self._ode_definition_map.values():
+            free_variable = ode.lhs.variables[0]
+            return free_variable
 
         raise ValueError('No free variable set in model.')  # pragma: no cover
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -314,18 +314,6 @@ class Model(object):
         """ Takes an RDF string and stores it in the model's RDF graph. """
         self.rdf.parse(StringIO(rdf))
 
-    def check_left_right_units_equal(self, equality):
-        """
-        Checks whether the LHS and RHS in a ``sympy.Eq`` have the same units.
-        :param equality: A ``sympy.Eq``.
-        """
-        lhs_units = self.units.evaluate_units(equality.lhs)
-        rhs_units = self.units.evaluate_units(equality.rhs)
-        assert self.units.is_equivalent(rhs_units, lhs_units), 'Units %s %s != %s %s' % (
-            self.units.format(lhs_units), self.units.format(lhs_units, True),
-            self.units.format(rhs_units), self.units.format(rhs_units, True),
-        )
-
     def get_equations_for(self, variables, recurse=True, strip_units=True):
         """Get all equations for a given collection of variables.
 
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -551,7 +551,6 @@ class Model(object):
     @property
     def graph(self):
         """ A ``networkx.DiGraph`` containing the model equations. """
-        # TODO: Set the parameters of the model (parameters rather than use initial values)
 
         # Return cached graph
         if self._graph is not None:
@@ -1143,7 +1142,6 @@ class VariableDummy(sympy.Dummy):
         self._set_cmeta_id(cmeta_id)
 
         # This variable's type
-        # TODO: Define allowed types via enum
         self.type = None
 
     def __str__(self):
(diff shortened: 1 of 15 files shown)


cellmlmanip/model.py
```diff
@@ -153,7 +182,7 @@ def add_number(self, value, units):
         Creates and returns a :class:`NumberDummy` to represent a number with units in sympy expressions.
 
         :param number: A number (anything convertible to float).
-        :param units: A string unit name or a ``Unit`` object.
+        :param units: A string unit name or a :class:`~cellmlmanip.units.UnitStore.Unit` object.
```
MichaelClerx, line 185: What's the ~ for?

cellmlmanip/model.py
```diff
@@ -311,30 +327,18 @@ def connect_variables(self, source_name: str, target_name: str):
         return True
 
     def add_rdf(self, rdf: str):
```
MichaelClerx, line 329: Are we using typing elsewhere?

cellmlmanip/model.py
```diff
@@ -536,8 +550,8 @@ def get_ontology_terms_by_variable(self, variable, namespace_uri=None):
     def get_definition(self, variable):
         """Get the equation (if any) defining the given variable.
 
-        :param variable: The variable to look up (as a class:`VariableDummy`. If this appears as the LHS of a straight
-            assignment, or the state variable in an ODE, the corresponding equation will be returned.
+        :param variable: The variable to look up (as a :class:`VariableDummy`). If this appears as the LHS of a
```
MichaelClerx, line 553: Did we not have a linewidth of 120 or something?

cellmlmanip/parser.py
```diff
@@ -485,6 +487,24 @@ def _parent_of(parent_name, child_name):
...
                          (comp_1, var_1, comp_2, var_2))
 
+    def transform_constants(self):
+        """
+        Standardise handling of 'constants'.
+
+        Once this has been called, the only variables with an initial_value attribute will be state variables,
+        and the initial value will do what it implies - hold the value the state variable should take at t=0.
+
+        Non state variables with an initial value are actually just constants. For consistent processing later on we add
```
MichaelClerx, line 497: "are treated as constants"?

cellmlmanip/units.py
```diff
@@ -1,4 +1,7 @@
-"""Unit handling for CellML models, using the Pint unit library."""
+"""
+The :mod:`cellmlmanip.units` module provides unit handling for CellML models, using the
+`Pint unit library <https://pint.readthedocs.io/>`_.
```
MichaelClerx, line 3: Should this be some other type of link, using the ones defined in conf.py?

docs/source/conf.py
```diff
@@ -67,15 +73,15 @@
...
+copyright = u'2019-, Cellmlmanip authors'
 author = u'Cellmlmanip authors'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = '0.0.1'
+version = '0.1.0'
```
MichaelClerx, line 84: :champagne:

docs/source/model_api.rst
```diff
@@ -2,6 +2,7 @@
 The Model API
 **************
 
-.. currentmodule:: cellmlmanip.model
 
-.. autoclass :: Model
+.. automodule:: cellmlmanip.model
+
+.. autofunction:: cellmlmanip.rdf.create_rdf_node
```
MichaelClerx, line 8: Is this the only `rdf` function that needs documenting? If so maybe it doesn't need its own module?

cellmlmanip/model.py
jonc125: It means it shows just as `Unit` in the link text, rather than the full dotted name.

cellmlmanip/model.py
jonc125: We were at one point, then stripped it out. Clearly this one was missed :)

cellmlmanip/model.py
jonc125: That was just me getting scared when my editor told me I was on column 121...

cellmlmanip/units.py
jonc125: It could be, but I think this kind is fine. Not sure what a special Sphinx link would be for the whole library?

docs/source/model_api.rst
jonc125: It is currently. But I think conceptually it's useful to keep it separate anyway? We don't have a utils module or similar.

PR #277, opened 2020-04-30
Unit conversion factors should have explicit units

jonc125:
## Description, motivation and Context

I spotted while doing #276 that `connect_variables` used a `NumberDummy` for the conversion factor but `convert_variable` did not. This PR updates `convert_variable` to address that anomaly.

As part of this, to make the new `NumberDummy` objects print nicely, I changed how their names are determined, which changed test results looking at the string value of expressions.

I've also made the typing of the value more consistent. The `_value` property now stores the value passed to the constructor, and conversion to `float` or `str` or `Float` is only done when calling `float(dummy)` or `str(dummy)` or `.evalf()` respectively.

The old `value` property has gone so users must be explicit about the type they want. If you were using it directly you'll need to update your code.

## Types of changes

- [x] Bug fix (**breaking** change which fixes an issue)

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -647,7 +647,7 @@ class Model(object):
 
             # Get all the dummy numbers on the RHS, and prepare substitution map
             dummies = equation.rhs.atoms(NumberDummy)
-            subs_dict = {d: sympy.Float(d.value, FLOAT_PRECISION) for d in dummies}
+            subs_dict = {d: d.evalf(FLOAT_PRECISION) for d in dummies}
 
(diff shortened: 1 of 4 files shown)


jonc125:
@MauriceHendrix given I managed to break your code unexpectedly with my last change, I won't merge this until you confirm it's OK 😁

MauriceHendrix:
> @MauriceHendrix given I managed to break your code unexpectedly with my last change, I won't merge this until you confirm it's OK 😁

it does "break" things but it's fine it's not a change that will break anything, just need to update the references.
Thanks for letting me know though it's much easier when I'm expecting it :D

Issue #278, opened 2020-05-03
Model.get_value(x) fails if x depends on variables


MichaelClerx:
Should just evaluate using the initial conditions

jonc125:
Fixed in #279.

PR #279, opened 2020-05-03
Made Model.get_value() evaluate states

MichaelClerx:
## Description
See #278 

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -312,8 +312,9 @@ class Model(object):
         Returns the evaluated value of the given variable, as a float.
 
         For state variables, this returns the initial value. For variables that depend on other variables this
-        recursively evaluates any dependencies (again at the initial state). Zero is returned for variables without a
-        definition (e.g. for the free variable).
+        recursively evaluates any dependencies (again at the initial state). Zero is returned for the free variable (as
+        identified by :meth:`get_free_variable()`); trying to evaluate other variables without a definition will
+        result in ``KeyError``.
         """
         return self._get_value(variable)
 
@@ -325,9 +326,7 @@ class Model(object):
             return float(variable.initial_value)
 
         # Get RHS and evaluate
-        expr = self._var_definition_map.get(variable)
-        if expr is None:
-            return 0
+        expr = self._var_definition_map[variable]
         expr = expr.rhs
         deps = expr.atoms(Variable)
         if deps:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -314,7 +314,7 @@ class Model(object):
         For state variables, this returns the initial value. For variables that depend on other variables this
         recursively evaluates any dependencies (again at the initial state). Zero is returned for the free variable (as
         identified by :meth:`get_free_variable()`); trying to evaluate other variables without a definition will
-        result in ``KeyError``.
+        result in a ``ValueError``.
         """
         return self._get_value(variable)
 
@@ -326,7 +326,12 @@ class Model(object):
             return float(variable.initial_value)
 
         # Get RHS and evaluate
-        expr = self._var_definition_map[variable]
+        try:
+            expr = self._var_definition_map[variable]
+        except KeyError:
+            if self._ode_definition_map and variable is self.get_free_variable():
+                return 0
+            raise ValueError('No definition set for ' + self.get_display_name(variable))
         expr = expr.rhs
         deps = expr.atoms(Variable)
         if deps:
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -325,7 +325,7 @@ class Model(object):
             return float(variable.initial_value)
 
         # Get RHS and evaluate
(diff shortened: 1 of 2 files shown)


cellmlmanip/model.py
```diff
@@ -560,8 +560,30 @@ def get_definition(self, variable):
...
+        deps = expr.atoms(VariableDummy)
+        while deps:
+            if evaluated is None:
+                evaluated = {x: x.initial_value for x in self._ode_definition_map.keys()}
+                expr = expr.xreplace(evaluated)
+            for dep in deps:
+                if dep not in evaluated:
+                    evaluated[dep] = self._get_value(dep, evaluated)
+            expr = expr.xreplace(evaluated)
+            deps = expr.atoms(VariableDummy)
```
jonc125, line 584: If `_get_value` is guaranteed to return a float, this should always return an empty set and you never loop. So I think we can get rid of the loop here.

cellmlmanip/model.py
```diff
@@ -560,8 +560,30 @@ def get_definition(self, variable):
...
+
+        return float(expr.evalf())
```
jonc125, line 586:
```suggestion
        return float(expr)
```
Conversion to float calls `evalf` as required.

tests/test_model_functions.py
```diff
@@ -333,9 +333,14 @@ def test_get_definition(self, simple_ode_model):
     def test_get_value(self, aslanidi_model):
         """ Tests Model.get_value() works correctly. """
 
-        symbol_a = aslanidi_model.get_variable_by_ontology_term((shared.OXMETA, "membrane_capacitance"))
+        # Test getting a constant
+        symbol_a = aslanidi_model.get_variable_by_ontology_term((shared.OXMETA, 'membrane_capacitance'))
         assert aslanidi_model.get_value(symbol_a) == 5e-5
 
+        # Test getting an intermediary value
```
jonc125, line 340: Is it worth adding a test that gets a state var directly?

cellmlmanip/model.py
```diff
@@ -560,8 +560,30 @@ def get_definition(self, variable):
         return defn
 
     def get_value(self, variable):
-        """Returns the evaluated value of the given variable's RHS."""
-        return float(self.graph.nodes[variable]['equation'].rhs)
+        """Returns the evaluated value of the given variable, as a float."""
```
jonc125, line 563: I think we should expand this docstring to emphasise that it recursively evaluates referenced equations and uses initial values for state variables.

We could also merge with the internal helper method, and document the extra parameter as for internal use only.

tests/test_model_functions.py
jonc125: Actually, we definitely should, as this case is uncovered!

cellmlmanip/model.py
MichaelClerx: Have updated the docstring, but much happier hiding the argument from the user! It's not part of the public API and also has some hidden constraints: If you were to insert an evaluation for `INa` that would be used, while an evaluation for `V` would not, and if you were to insert evaluations for both states and intermediary values, you'd have to be sure those were consistent or the method behaviour would be undefined

cellmlmanip/model.py
```diff
@@ -308,8 +308,38 @@ def get_definition(self, variable):
...
+
+    def _get_value(self, variable, evaluated=None):
+        """Internal method to implement :meth:`get_value()`."""
+
+        # State? Then return initial value
+        if variable in self._ode_definition_map:
+            return float(variable.initial_value)
+
+        # Get RHS and evaluate
+        expr = self.graph.nodes[variable]['equation']
```
jonc125, line 328: Ooh, actually, we should use
```suggestion
        expr = self._var_definition_map.get(variable)
```
for this now!

cellmlmanip/model.py
```diff
@@ -308,8 +308,38 @@ def get_definition(self, variable):
...
+        if expr is None:
+            return 0
```
jonc125, line 330: Wouldn't it be better to raise a sensible error for this? More like an undefined variable case isn't it?

jonc125: Ah, I see from the test below this is to support time. Probably better to handle time explicitly; see suggestion below?

cellmlmanip/model.py
```diff
@@ -308,8 +308,38 @@ def get_definition(self, variable):
...
+        expr = expr.rhs
+        deps = expr.atoms(VariableDummy)
+        if deps:
+            if evaluated is None:
+                evaluated = {x: x.initial_value for x in self._ode_definition_map.keys()}
```
jonc125, line 335:
```suggestion
                evaluated = {x: x.initial_value for x in self._ode_definition_map.keys()}
                if self._ode_definition_map:
                    time = self.get_free_variable()
                    evaluated[time] = 0
```

MichaelClerx:
Sorry did another slight update @jonc125 :D

PR #281, opened 2020-05-04
Rename our dummy classes to better reflect their purpose

### LINKED ISSUE #161: Rename/move some Model functions

jonc125:
## Description
- `VariableDummy -> Variable`
- `NumberDummy -> Quantity`
- `Model.add_number -> Model.create_quantity`

## Motivation and Context
Fixes #161. Means the names better reflect what they are for, rather than internal Sympy details.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -32,7 +32,7 @@ class Model(object):
     graph that stores further meta data about the model.
 
     Equations are stored as :class:`sympy.Eq` objects, but with the caveat that all variables and numbers must be
-    specified using the :class:`sympy.Dummy` objects returned by :meth:`add_variable()` and :meth:`add_number()`.
+    specified using the :class:`sympy.Dummy` objects returned by :meth:`add_variable()` and :meth:`create_quantity()`.
 
     Units are handled using the ``units`` property of a model, which is an instance of
     :class:`cellmlmanip.units.UnitStore`.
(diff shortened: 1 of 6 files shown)


cellmlmanip/model.py
```diff
@@ -1064,26 +1067,26 @@ def _convert_variable_instance(self, original_variable, cf, units, direction):
...
+    You can also use ``quantity.evalf()`` to get the value as a :class:`sympy.Float`.
     """
 
     # Sympy annoyingly overwrites __new__
     def __new__(cls, value, *args, **kwargs):
         # Middle argument is the symbol name, so must be a string
         if not isinstance(value, str):
             value = '{:g}'.format(value)
-        return super().__new__(cls, value, real=True)
+        return super().__new__(cls, '_' + value, real=True)
```
jonc125, line 1089: I have no idea why sympy stopped adding the underscore itself when I renamed the class, but apparently it did!

MauriceHendrix: that seems off you sure it's not a different version of sympy?

jonc125: I don't think so - I didn't change any library versions locally, only change was this PR itself.

Issue #208, opened 2020-02-03
Speed up model parsing


MichaelClerx:
Currently this is very very slow. For example, Myokit is approx ~30 times faster (only checked one file). We should do some profiling to see what causes this, and then check if there's some easy fixes to speed things up

jonc125:
@MauriceHendrix would you be up for doing some profiling on e.g. regenerating the Chaste codegen references, so we have some data to guide optimisation?

I have a couple of ideas for improving the performance of `get_equations_for` specifically, but don't want to invest time in them unless it's going to be noticed!

1. Cache the result of `sorted_variables = nx.lexicographical_topological_sort(graph, key=str)` for each graph, since that doesn't change unless the graph does. Does this method take a long time?
2. Change the logic in `get_equations_for` so it doesn't do a sort, instead working backwards along graph edges from the nodes supplied to build a list of just the equations actually needed. It's possible this would change the ordering in generated code, but I think it should be a fair bit faster as it builds fewer interim data structures.

MauriceHendrix:
sure would be interested. Also in codegen in a few places I add stuff to the model and than just use get_equatons_for again. It would probably be faster if I just updated the list I already had, but that would be quite a bit of extra work.

MauriceHendrix:
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    75362    0.043    0.000    0.124    0.000 <frozen importlib._bootstrap>:1009(_handle_fromlist)
427730/34403    0.205    0.000    5.535    0.000 assumptions.py:452(getit)
150408/2785    0.716    0.000    5.503    0.002 assumptions.py:464(_ask)
    17634    0.021    0.000    0.026    0.000 basic.py:108(__new__)
36206/36148    0.019    0.000    5.713    0.000 basic.py:1106(xreplace)
658184/36148    1.197    0.000    5.693    0.000 basic.py:1171(_xreplace)
   886729    0.210    0.000    0.210    0.000 basic.py:738(args)
726569/708497    0.208    0.000    0.228    0.000 expr.py:106(__hash__)
    13102    0.154    0.000    0.327    0.000 facts.py:499(deduce_all_facts)
  2200631    1.474    0.000    2.120    0.000 random.py:224(_randbelow)
   138250    1.017    0.000    3.153    0.000 random.py:264(shuffle)
701088/700836    0.197    0.000    0.275    0.000 {built-in method builtins.getattr}
[...]

PR #284, opened 2020-05-06
Improving MathML transpiling

### LINKED ISSUE #51: Rewrite transpiler to use etree instead of dom
Investigate why the mathml2sympy transpiler doesn't like a reference to the MathML DOM element. Currently passing string representation of the MathML.

See `_add_maths` in `parser.Parser`

jonc125:
## Description
Moved the transpiler into parser.py and made it use lxml (elementtree API) rather than DOM. This means we don't need to serialise MathML to string & re-parse to convert to sympy.

## Motivation and Context
Fixes #51. Should also help with #208.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue) - at least, it's non-breaking provided users haven't accessed transpiler.py directly!

## Checklist:

## Testing


tests/test_model_evaluation.py
```diff
@@ -1,90 +1,76 @@
...
-
 import networkx
 import pytest
 import sympy
 
-from cellmlmanip import load_model
-
 
-class TestModelEvaluation():
+def test_model_evaluation(hh_model):
```
MichaelClerx, line 6: This could probably go in somewhere else then? Or maybe leave for a bigger test reorganisation :-)

MichaelClerx:
> Also (but maybe less pressing), I found that if you use an assert instead of raising a value error it plays nicely with the code coverage, so e.g.
> `assert len(result) == 2, 'Need exactly 2 children for <piece>'`
> which f triggered gives:
> `AssertionError: Need exactly 2 children for <piece>`
> as opposed to

Asserts should never go from cellmlmanip to a cellmlmanip user though. They're sanity checks, not part of the the public API.

MichaelClerx:
Also, coverage and style checks are there to help. So adapting the code for no other reason than to make them pass is a terrible idea :D

MichaelClerx:
I even try and write commit messages like "improved code style" rather than "fixed flake8", just to remind myself that flake8 is a tool, not the goal :-|
(sorry for being so boring)

MauriceHendrix:
> Also, coverage and style checks are there to help. So adapting the code for no other reason than to make them pass is a terrible idea :D

Sure, but really in my opinion an `assert` is a better fit for something that should be true but where you think it feels like something that might go wrong in the future, even if there is no way of making it fail right now. 
I would say `raise` is more appropriate for errors you are actively expecting and handling.

MichaelClerx:
This one's asking how many children in a `<piece>` though, that's determined by the CellML file you give it as input, isn't it? Might be that this gets caught by an earlier validation, which would make it unreachable and in that case I fully agree with you :D

MauriceHendrix:
> This one's asking how many children in a `<piece>` though, that's determined by the CellML file you give it as input, isn't it? Might be that this gets caught by an earlier validation, which would make it unreachable and in that case I fully agree with you :D

Well if it doesn't then it would need a test contrived cellml file as case no?

PR #287, opened 2020-05-18
Fc data interpolation

jonc125:
## Description
* More robust support for algebraic models. In particular:
    * `get_free_variable` is now covered in our unit tests for this case
    * `convert_variable` will now work for an algebraic model
* Added `move_annotations` optional argument to `convert_variable`, so that we don't have to transfer the cmeta:id in all cases. I needed a 'read-only' converted version of an annotated variable that didn't change the canonical version referred to by metadata.

## Motivation and Context
These features were needed to implement https://github.com/ModellingWebLab/weblab-fc/issues/109.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/cellmlmanip/model.py
+++ b/cellmlmanip/model.py
@@ -714,7 +714,7 @@ class Model(object):
         # Update mapping
         self._cmeta_id_to_variable[target._cmeta_id] = target
 
-    def convert_variable(self, original_variable, units, direction):
+    def convert_variable(self, original_variable, units, direction, move_annotations=True):
         """
         Ensures the model contains a variable representing ``original_variable`` in the specified ``units``.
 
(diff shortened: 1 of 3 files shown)


cellmlmanip/model.py
```diff
@@ -97,7 +97,7 @@ def get_free_variable(self):
             free_variable = ode.lhs.variables[0]
             return free_variable
 
-        raise ValueError('No free variable set in model.')  # pragma: no cover
+        raise ValueError('No free variable set in model.')
```
MichaelClerx, line 100: Not sure what this exception should be. We're not really complaining about the value of a passed parameter here. Maybe a custom type?

tests/test_parser.py
```diff
@@ -246,14 +246,18 @@ def test_bad_connection_units(self):
 
         assert match in str(dim_error.value)
 
-    def test_new_base_units(self):
-        """Tests unit checking in a model that defines a new base unit."""
+    def test_algebraic_and_new_base_units(self):
```
MichaelClerx, line 249: Can we split this in two tests? These are two entirely unrelated things

cellmlmanip/model.py
jonc125: I was thinking of it as the value of the object you're calling it on being inappropriate for the method. So it's the `self` parameter it's complaining about? It's not really a TypeError, since the type is correct.

tests/test_parser.py
jonc125: We could, but keep the same CellML file?

MichaelClerx: sure

PR #296, opened 2020-06-08
Update setup.py

MauriceHendrix:
## Description

I updated setup.py

## Motivation and Context

test Pypi  didn't like the way it was trying to read the license and it repented the license to the description and quoted the (markdown) description, not a good look. This version shows up better

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,18 +4,21 @@ from setuptools import find_packages, setup
 with open('README.md') as f:
     readme = f.read()
 
-with open('LICENSE') as f:
-    license_ = f.read()
-
 setup(
     name='cellmlmanip',
     version='0.1.0',
     description='CellML loading and model equation manipulation',
     long_description=readme,
+    long_description_content_type="text/markdown",
     author='Asif U Tamuri, Sarah M Keating, Maurice Hendrix, Michael Clerx, Jonathan Cooper',
     author_email='j.p.cooper@ucl.ac.uk',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -4,18 +4,21 @@
...
 
-with open('LICENSE') as f:
-    license_ = f.read()
-
 setup(
     name='cellmlmanip',
     version='0.1.0',
     description='CellML loading and model equation manipulation',
     long_description=readme,
+    long_description_content_type="text/markdown",
```
MichaelClerx, line 12: We're using `'` in the rest of the arguments

MauriceHendrix: sure

setup.py
```diff
@@ -4,18 +4,21 @@
...
     version='0.1.0',
     description='CellML loading and model equation manipulation',
     long_description=readme,
+    long_description_content_type='text/markdown',
     author='Asif U Tamuri, Sarah M Keating, Maurice Hendrix, Michael Clerx, Jonathan Cooper',
     author_email='j.p.cooper@ucl.ac.uk',
     url='https://github.com/ModellingWebLab/cellmlmanip',
-    license=license_,
+    classifiers=[
+        "Programming Language :: Python :: 3",
```
MichaelClerx, line 17: And here

MauriceHendrix: yes but are you ok with the change in principle?

PR #302, opened 2020-06-23
Optional unsorted results

MauriceHendrix:
## Description

Allow unsorted results for the get_state_variables, get_derivatives, get_derived_quantities and get_variables_by_rdf methods of Model

## Motivation and Context

This would be useful for codegen and maybe other components where I get the result of these calls and either sort it myself (in codegen a specific order) or feed it into a set where the order is obviously not important.

## Types of changes

Options for skipping sorting to improve performance

## Documentation checklist

## Testing


setup.py
```diff
@@ -37,7 +37,7 @@
             'codecov',
             'flake8',
             'isort',
-            'pytest>=3.2',
+            'pytest>=4.6',
```
MauriceHendrix, line 40: bumped version as otherwise non-pinned tests will fail

cellmlmanip/model.py
```diff
@@ -232,10 +245,11 @@ def get_rdf_value(self, subject, predicate):
 
         Note: expects exactly one triple to match and the result to be a literal. Its string value is returned.
         """
-        triples = list(self.get_rdf_annotations(subject, predicate))
-        assert len(triples) == 1
-        assert isinstance(triples[0][2], rdflib.Literal)
-        value = str(triples[0][2]).strip()  # Could make this cleverer by considering data type if desired
+        triples = self.get_rdf_annotations(subject, predicate)
+        triple = next(triples)
```
MichaelClerx, line 249: Not sure why these changes are in this PR. But this will raise an iterator error if triples is empty, no? So I think `assert len(triples) == 1, 'Expecting exactly 1 triple'` would be better?

cellmlmanip/model.py
```diff
@@ -502,7 +516,7 @@ def graph_with_sympy_numbers(self):
                 # Check if simplification removed dependencies on other variables, and if so remove the corresponding
                 # edges.
                 refs = self.find_variables_and_derivatives([rhs])
-                edges = list(graph.in_edges(equation.lhs))
+                edges = tuple(graph.in_edges(equation.lhs))
```
MichaelClerx, line 519: Is this just for speed?

MauriceHendrix: yup iterating over immutable (i.e. tuples) or sets is quite a faster than over lists

cellmlmanip/model.py
MauriceHendrix: Good point I changed it to.
```
        triple = next(triples, None)
        assert triple is not None, 'Expecting exactly 1 triple'
```
To catch that case it's just that I want to avoid unnecessary list creation.

MichaelClerx:
```
>>> next(iter([]))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
StopIteration
```

MichaelClerx: Ah ok. Just spotted the `None`

MichaelClerx: Hmmmm, what about `triples = [None]`? That triggers the same assert?

MauriceHendrix: I just checked rdflib

RELEASE.md
```diff
@@ -1,2 +1,6 @@
+# Release 0.2.0
+It is now possible to pass sort=False to the get_state_variables, get_derivatives, get_derived_quantities and get_variables_by_rdf methods of Model to get unsorted results.
```
jonc125, line 2: Minor suggestion: let's use the Markdown format to make the release notes look nice. I suggest a bullet for each change, and use monospace for code fragments, eg `sort=False`, `Model`

cellmlmanip/model.py
```diff
@@ -232,10 +245,12 @@ def get_rdf_value(self, subject, predicate):
...
-        triples = list(self.get_rdf_annotations(subject, predicate))
-        assert len(triples) == 1
-        assert isinstance(triples[0][2], rdflib.Literal)
-        value = str(triples[0][2]).strip()  # Could make this cleverer by considering data type if desired
+        triples = self.get_rdf_annotations(subject, predicate)
+        triple = next(triples, None)
+        assert triple is not None, 'Expecting exactly 1 triple'
+        assert next(triples, None) is None, "Expecting exactly 1 triple"
+        assert isinstance(triple[2], rdflib.Literal)
+        value = str(triple[2]).strip()  # Could make this cleverer by considering data type if desired
```
jonc125, line 253: This change is presumably such a minor speed optimisation that I'd be inclined to leave it as it was, since that was clearer.

tests/test_model_functions.py
```diff
@@ -65,9 +65,13 @@ def test_get_derived_quantities(self, basic_model, simple_ode_model):
         """ Tests Model.get_derived_quantities(). """
 
         derived_quantities = basic_model.get_derived_quantities()
+        unsorted_derived_quantities = basic_model.get_derived_quantities(sort=False)
+        assert set(derived_quantities) == set(unsorted_derived_quantities)
```
jonc125, line 69: Given it's an empty set testing the sorting is a bit pointless :)

tests/test_model_functions.py
```diff
@@ -116,12 +120,16 @@ def test_get_state_variables(self, basic_model):
         """ Tests Model.get_state_variables() works on a simple model. """
 
         states = basic_model.get_state_variables()
+        unsorted_states = basic_model.get_state_variables(sort=False)
+        assert set(states) == set(unsorted_states)
```
jonc125, line 124:
```suggestion
        assert states == unsorted_states
```
Since there's just one!

cellmlmanip/model.py
MauriceHendrix: hhm ok

PR #303, opened 2020-06-23
Parser speed optimisation

MauriceHendrix:
## Description

Improved parser speed by delaying adding mathematics until connections are known.

I also added tests for them, and in a few places in the parser changed the error thrown to and assert as I didn't I couldn't come up with a way to make it every throw that error, let me know if I've missed something there.

## Motivation and Context

Parsing wasn't as fast as it could be

## Types of changes

Speed improvement

## Documentation checklist

## Testing


cellmlmanip/model.py
```diff
@@ -936,15 +936,16 @@ def connect_variables(self, source_name, target_name):
...
-            for index, equation in enumerate(self.equations):
-                self.equations[index] = new_eq = equation.xreplace({target: source.assigned_to})
-                if equation.lhs.is_Derivative:
-                    state_var = equation.lhs.free_symbols.pop()
-                    assert state_var in self._ode_definition_map
-                    self._ode_definition_map[state_var] = new_eq
-                else:
-                    assert equation.lhs in self._var_definition_map
-                    self._var_definition_map[equation.lhs] = new_eq
+            if len(self.equations) > 0:
```
MauriceHendrix, line 939: my timing tests showed that it's faster to check len(dict) >0 than to enumerate over an empty dict.
Since 99% of the use is from the parser, I introduced this as a speed-up, I could have removed it entirely, but then we get a library method that doesn't quite do what you would expect, so that's why I kept it.

cellmlmanip/parser.py
```diff
@@ -68,22 +68,19 @@ def __init__(self, name):
...
-        if sibling_name in self.siblings:
-            raise ValueError('Sibling component %s already added!' % sibling_name)
+        assert sibling_name not in self.siblings, 'Sibling component %s already added!' % sibling_name
         self.siblings.add(sibling_name)
 
     def add_encapsulated(self, encapsulated_name):
         """Adds an encapsulated component to this component"""
-        if encapsulated_name in self.encapsulated:
-            raise ValueError('Encapsulated component %s already added!' % encapsulated_name)
+        assert encapsulated_name not in self.encapsulated, 'Encapsulated component %s already added!' % \
```
MauriceHendrix, line 82: I really can't come up with a test for this since the `_Component` class is private and I don't think it should ever be possible, hence I turned it into an assert (with the added benefit that it doesn't count as a coverage miss). Let me know what you think.

setup.py
```diff
@@ -37,7 +37,7 @@
             'codecov',
             'flake8',
             'isort',
-            'pytest>=3.2',
+            'pytest>=4.6',
```
MauriceHendrix, line 40: bumped version as otherwise non-pinned tests will fail

cellmlmanip/model.py
```diff
@@ -834,8 +834,7 @@ def convert_variable(self, original_variable, units, direction, move_annotations
         if original_variable == free_symbol:
             # Change every ODE to be w.r.t. the new time variable
             # Process ODEs in model order to ensure new equations are added in a consistent order
-            for ode in [eq for v, eq in sorted(self._ode_definition_map.items(),
-                                               key=lambda v_eq: v_eq[0].order_added)]:
+            for _, ode in sorted(self._ode_definition_map.items(), key=lambda v_eq: v_eq[0].order_added):
```
MauriceHendrix, line 837: No need for extra list creation

cellmlmanip/model.py
MichaelClerx: How much does this save though?

cellmlmanip/parser.py
```diff
@@ -432,10 +444,11 @@ def _add_connection(self, model):
                 connections_to_process.append(connection)
                 unchanged_loop_count += 1
             else:
+                connected_variable_mapping[str(connection[1])] = self.model.get_variable_by_name(connection[0])
                 unchanged_loop_count = 0
 
-            if unchanged_loop_count > len(connections_to_process):
-                raise ValueError('Unable to add connections to the model')
+            assert unchanged_loop_count <= len(connections_to_process), 'Unable to add connections to the model'
```
MichaelClerx, line 450: I don't think this should be an assert? Looks to me it can go wrong depending on the model that's being parsed, so this is an error for the user (Exception) not for the cellmlmanip developer (assert)

cellmlmanip/parser.py
```diff
@@ -310,41 +312,49 @@ def _add_variables(self, component_element):
 
         return variable_lookup_symbol
 
-    def _add_maths(self, component_element, variable_to_symbol):
+    def _add_maths(self, maths_to_add, connected_variable_mapping):
         """
+        Add maths for all elements sorted in maths_to_add.
```
MichaelClerx, line 317: "sorted in maths_to_add" ?

cellmlmanip/parser.py
```diff
@@ -310,41 +312,49 @@ def _add_variables(self, component_element):
...
+
         <model> <component> <math> </component> </model>
 
-        :param component_element: an etree.Element
-        :param variable_to_symbol: a ``Dict[str, sympy.Dummy]``
+        :param maths_to_add: a list of (component_element, variable_to_symbol) tuples
```
MichaelClerx, line 321: Great docs!

cellmlmanip/parser.py
```diff
@@ -310,41 +312,49 @@ def _add_variables(self, component_element):
...
+                             where ``component_element`` is an ``etree.Element`
+                             and ``variable_to_symbol`` is a ``Dict[str, sympy.Dummy]``
+
+        :param connected_variable_mapping: a ``Dict[str, sympy.Dummy]`` mapping a connected variable to its source.
```
MichaelClerx, line 325: Wondering about `maths_to_add` as a variable name: I was expecting it to contains maths (e.g. maths elements or sympy equations), but it contains component elements. Looking through the code I think you're right that this is the most efficient way to do it, but maybe the name should be changed?

cellmlmanip/parser.py
MichaelClerx: I guess it was trying to capture the case where the CellML specifies the same encapsulation relationship twice?
If so, you should be able to trigger it with something like

```
   <group>
      <relationship_ref relationship="encapsulation"/>
      <component_ref component="sodium_current">
         <component_ref component="sodium_current_m_gate"/>
         <component_ref component="sodium_current_m_gate"/>
      </component_ref>
   </group>
```
and/or
```
   <group>
      <relationship_ref relationship="encapsulation"/>
      <component_ref component="sodium_current">
         <component_ref component="sodium_current_m_gate"/>
      </component_ref>
   </group>
   <group>
      <relationship_ref relationship="encapsulation"/>
      <component_ref component="sodium_current">
         <component_ref component="sodium_current_m_gate"/>
      </component_ref>
   </group>
```
(the second one is not allowed for 2 reasons, so might raise an error somewhere else)

cellmlmanip/model.py
MauriceHendrix: On my laptop time difference: 3.0e-07s vs  9.3e-08s in the example model we're calling this 81 times so it would save 
1.7e-05s so not really noticable to be fair. Come to think of it it doesn't seem to be used anywhere but the parser so maybe it's better to remove it altogether.

cellmlmanip/parser.py
MauriceHendrix: No

jonc125: Ah, Asif cutting corners then :)

cellmlmanip/parser.py
MauriceHendrix: stored, sorry

cellmlmanip/parser.py
MauriceHendrix: @MichaelClerx let me know if you have any constructive feedback thanks!

cellmlmanip/parser.py
```diff
@@ -310,41 +312,49 @@ def _add_variables(self, component_element):
...
+                        out = connected_variable_mapping[str(out)]
+                    assert out is not None, '%s not found in symbol dict' % (prefix + identifer)
+                    return out
 
-        # reuse transpiler so dummy symbols are kept across <math> elements
-        transpiler = Transpiler(
-            symbol_generator=symbol_generator,
-            number_generator=lambda x, y: self.model.create_quantity(x, self.model.units.get_unit(y)),
-        )
+                # reuse transpiler so dummy symbols are kept across <math> elements
```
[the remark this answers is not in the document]
MauriceHendrix: maybe not anymore more useful comment

cellmlmanip/parser.py
```diff
@@ -132,9 +129,11 @@ def parse(self, unit_store=None):
         self._add_units(model_xml)
         self._add_rdf(model_xml)
 
-        self._add_components(model_xml)
+        maths_to_add = self._add_components(model_xml)
         self._add_relationships(model_xml)
-        self._add_connection(model_xml)
+
+        connected_variable_mapping = self._add_connection(model_xml)
```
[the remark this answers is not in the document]
MauriceHendrix: but it's not a function that adds connections, it's a dictionary that keeps a mapping of what has been connected

cellmlmanip/parser.py
MauriceHendrix: you're right of course better like this?

cellmlmanip/parser.py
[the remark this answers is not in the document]
jonc125: `_process_connections` then? But definitely not connection singular :)

tests/test_model_functions.py
```diff
@@ -673,47 +673,6 @@ def test_find_variables_and_derivatives_2(self, hh_model):
         assert t in syms
         assert sp.Derivative(v, t) in syms
 
-    def test_connect_variables(self, local_hh_model):
```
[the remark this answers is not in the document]
MauriceHendrix: it's covered in test_connections in test_parser.py and in various other bits (as an aside)

cellmlmanip/parser.py
MauriceHendrix: I see ok test added now

tests/cellml_files/test_simple_odes.cellml
```diff
@@ -52,7 +52,7 @@ For annotations, we check:
       <unit units="second" prefix="micro"/>
    </units>
    <units name="mV">
-      <unit units="volt" prefix="milli"/>
+      <unit units="volt" prefix="-3"/>
```
MauriceHendrix, line 55: Just to trigger the part of the code in the parser test that deals with prefixes that are numbers it's also already covered elsewhere

cellmlmanip/parser.py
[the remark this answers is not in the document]
MauriceHendrix: _add_connections is fine I thought you meant the line below sorry

docs/source/conf.py
```diff
@@ -19,6 +19,8 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 import cellmlmanip
+
```
MauriceHendrix, line 22: Isort suddenly insisted on this

cellmlmanip/parser.py
MichaelClerx: It was genuine, I was pleased to see it!

cellmlmanip/parser.py
```diff
@@ -68,16 +68,13 @@ def __init__(self, name):
 
     def set_parent(self, parent_name):
         """Sets the parent of this component"""
-        if self.parent:
-            raise ValueError('Parent of component %s already %s. Cannot set %s!' % (self.name,
-                                                                                    self.parent,
-                                                                                    parent_name))
+        assert not self.parent, 'Parent of component %s already %s. Cannot set %s!' % (self.name, self.parent,
```
MichaelClerx, line 71: Do we want an assert here? I think it depends on whether the CellML is pre-validated or not. If the CellML is valid, this should never happen, so an assert is fine. If calling this method is part of how we validate the CellML, then this should be a user error, that says something like "Multiple parents specified for component X".
Same for `add_sibling` below.

What do you think, @jonc125 ?

MauriceHendrix: @MichaelClerx  Ok actually I finally found a way to trigger this with wrong cellml, so I'll reinstate the raise error and add a test

MauriceHendrix: @MauriceHendrix  Ok I found a way to trigger this, I'll revert + test

PR #304, opened 2020-06-23
cache sorting in get_equations_for

MauriceHendrix:
## Description

Cache the graph sorting in get_equations_for

## Motivation and Context

Sorting takes time, so this is a (small) speed improvement

## Types of changes

Speed improvement

## Documentation checklist

## Testing


jonc125:
I'd agree that we want some numbers in the PR description to show what level of speed-up this gives.

MauriceHendrix:
> I'd agree that we want some numbers in the PR description to show what level of speed-up this gives.

Hhm I just did a time and it was marginally slower with caching. I don't really understand but probably an indication it's not a change we want.

PR #306, opened 2020-06-24
bumped version and moved version to 1 place

MauriceHendrix:
## Description

- Bumped version in preparation for new release.
- Moved version number to version.txt which is read from setup.py and docs/config.py
- Also added functionality to cellmlmanip to report its own version via version() and __version__ + tests
- Also bumped pytest version

## Motivation and Context

- If the version number is just in 1 place, there is less of a risk of forgetting to bump especially the docs version number.
- Pytest version bump was needed to make non-pinned tests pass.

## Types of changes

## Documentation checklist

## Testing


cellmlmanip/_config.py
```diff
@@ -0,0 +1,40 @@
...
+# Expose version number
+#
+def version(formatted=False):
+    """
+    Returns the version number, as a 3-part integer (major, minor, revision).
+    If ``formatted=True``, it returns a string formatted version (e.g.
+    "cellmlmanip 1.0.0").
+    """
+    if formatted:
+        return 'chaste_codegen ' + __version__
```
jonc125, line 38:
```suggestion
        return 'cellmlmanip ' + __version__
```

docs/source/conf.py
```diff
@@ -81,7 +83,8 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.1.0'
+with open(os.path.join('..', 'cellmlmanip', 'version.txt'), 'r') as f:
```
jonc125, line 86: I think the documentation has to import the cellmlmanip library elsewhere, so might be better to do the same here?

setup.py
```diff
@@ -37,7 +44,7 @@
             'codecov',
             'flake8',
             'isort',
-            'pytest>=3.2',
+            'pytest>=4.6',
```
jonc125, line 47: What has changed that this is needed?

tests/test_version.py
```diff
@@ -0,0 +1,29 @@
...
+    assert isinstance(version[1], int)
+    assert isinstance(version[2], int)
+    assert version[0] >= 0
+    assert version[1] >= 0
+    assert version[2] >= 0
+
+    version = cellmlmanip.version(formatted=True)
+    assert isinstance(version, str)
+    assert len(version) >= 1
+    assert version.startswith('chaste_codegen ')
```
jonc125, line 29:
```suggestion
    assert version.startswith('cellmlmanip ')
```

setup.py
```diff
@@ -1,12 +1,19 @@
...
+
 from setuptools import find_packages, setup
 
 
+# Load README
 with open('README.md') as f:
     readme = f.read()
 
+# Load version number
+with open(os.path.join('cellmlmanip', 'version.txt'), 'r') as f:
```
jonc125, line 11: Have you checked that the version.txt file is installed as part of cellmlmanip?

cellmlmanip/_config.py
```diff
@@ -0,0 +1,40 @@
...
+    MODULE_DIR = os.path.dirname(inspect.getfile(frame))
+finally:
+    # Always manually delete frame
+    # https://docs.python.org/2/library/inspect.html#the-interpreter-stack
+    del(frame)
+
+#
+# Version info
+#
+with open(os.path.join(MODULE_DIR, 'version.txt'), 'r') as f:
```
jonc125, line 23: Is this the recommended way of approaching this problem? Or would [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access) be better?

setup.py
MauriceHendrix: as you know the pinned tests fail if we don't do this, same as with codegen, it's due to the latest pytest-cov (which is the line below it) now requiring a higher pytest version.

setup.py
MauriceHendrix: uups now I have!

cellmlmanip/_config.py
MauriceHendrix: it's one of the recommendations on here:
https://packaging.python.org/guides/single-sourcing-package-version/
Since the docs conf.py doesn't (directly) import cellmlmanip, pkg_resources  probably wouldn't work but the version.txt approach should

docs/source/conf.py
MauriceHendrix: it doesn't import the cellmlmanip library

jonc125: I think Sphinx does in order to generate the docs for it though, from the docstrings.

tests/test_version.py
```diff
@@ -0,0 +1,29 @@
+#
+# Tests the basics of chaste_codegen
```
MichaelClerx, line 2: No it doesn't :D

docs/source/conf.py
MichaelClerx: I think it would be better to just `import cellmlmanip; version = release = cellmlmanip.__version__` here

tests/test_version.py
MauriceHendrix: uups I sometimes don't know which file I'm looking at :D

docs/source/conf.py
MichaelClerx: This one!

MauriceHendrix: ok as @MichaelClerx  and @jonc125 want can you please approve it?

PR #310, opened 2020-06-30
Codecov improvements

MauriceHendrix:
## Description

Improvements to test coverage

## Motivation and Context

Better etst coverage may be nice to have for release

## Types of changes

Code coverage improvements

## Documentation checklist

I don't think this needs to be in release notes?

## Testing


cellmlmanip/parser.py
```diff
@@ -75,8 +75,7 @@ def set_parent(self, parent_name):
 
     def add_sibling(self, sibling_name):
         """Adds a sibling for this component"""
-        if sibling_name in self.siblings:
-            raise ValueError('Sibling component %s already added!' % sibling_name)
+        assert sibling_name not in self.siblings, 'Sibling component %s already added!' % sibling_name
```
MauriceHendrix, line 78: This doesn't seem to be reachable, when I tried to create it I got earlier errors about the variable already being assigned to.

cellmlmanip/printer.py
```diff
@@ -288,9 +280,9 @@ def _print_Mul(self, expr):
 
         # Fix brackets for Pow with exp -1 with more than one Symbol
         for item in pow_brackets:
-            if item.base in b:
-                b_str[b.index(item.base)] = \
-                    '(' + b_str[b.index(item.base)] + ')'
+            assert item.base in b, "item.base should be kept in b for powers"
```
MauriceHendrix, line 283: I couldn't trigger this I don't think it is/ should be possible.

cellmlmanip/units.py
```diff
@@ -809,19 +806,6 @@ def add_context(self, expression, message):
         self.message += ('. Context: ' + message).format(expression)
 
 
-class UnitsCannotBeCalculatedError(UnitError):
```
MauriceHendrix, line 812: This one was no longer in use

tests/test_printer.py
```diff
@@ -267,3 +272,11 @@ def test_print_model_equations(model, simple_ode_model, p):
              "_units_conversion1$sv1, _environment$time)', 'Derivative(_state_units_conversion2$sv1, _environment$time"
              ") = 1000.0', 'deriv_on_rhs2b$sv1_rate = Derivative(_state_units_conversion2$sv1, _environment$time)', 's"
              "ingle_ode_rhs_const_var$a = 1.0']")
+
+    def test_pow_non_commutative(self, p, x, y):
```
MauriceHendrix, line 276: Extra test to hit previously untested case.

cellmlmanip/units.py
```diff
@@ -267,9 +267,6 @@ def get_conversion_factor(self, from_unit, to_unit):
         cf = self.convert(1 * from_unit, to_unit).magnitude
         if isinstance(cf, numbers.Number) and math.isclose(cf, 1.0):
             return 1
-        elif isinstance(cf, sympy.Mul) and 1.0 in cf.args:
```
MauriceHendrix, line 270: This code wasn't being triggered by the tests, seems not to be necessary anymore.

cellmlmanip/parser.py
MichaelClerx: I confirmed, like this:

```

    def parse(self, cellml, wrap=True):
        """Writes some CellML to a temporary file and parses it."""
        if wrap:
            cellml = (
                '<?xml version="1.0" encoding="UTF-8"?>'
                '<model name="test" xmlns="http://www.cellml.org/cellml/1.0#">'
                + cellml +
                '</model>'
            )

        with tempfile.TemporaryDirectory() as d:
            path = os.path.join(d, 'file.cellml')
            with open(path, 'w') as f:
                f.write(cellml)
            return parser.Parser(path).parse()

    def test_encapsulation(self):
        """ Tests """

        # Parse valid sibling relationship
        self.parse(
            '<component name="a" />'
            '<component name="b" />'
            '<component name="c" />'
            '<group>'
            ' <relationship_ref relationship="encapsulation"/>'
            ' <component_ref component="a">'
            '  <component_ref component="b" />'
            '  <component_ref component="c" />'
            ' </component_ref>'
            '</group>'
        )
        # Note: this isn't stored, so can't check if it worked via public API!

        # Parse invalid relationship
        self.parse(
            '<component name="a" />'
            '<component name="b" />'
            '<component name="c" />'
            '<group>'
            ' <relationship_ref relationship="encapsulation"/>'
            ' <component_ref component="a">'
            '  <component_ref component="b" />'
            '  <component_ref component="c" />'
            ' </component_ref>'
            '</group>'
            '<group>'
            ' <relationship_ref relationship="encapsulation"/>'
            ' <component_ref component="a">'
            '  <component_ref component="b" />'
            '  <component_ref component="c" />'
            ' </component_ref>'
            '</group>'
        )
```

:D 

E           ValueError: Encapsulated component b already added!

cellmlmanip/units.py
```diff
@@ -411,8 +411,6 @@ def _check_unit_of_quantities_equal(self, list_of_quantities):
         :param list_of_quantities: a list of ``pint.Quantity`` objects.
         :returns: boolean indicating whether all units are equivalent
         """
-        if len(list_of_quantities) == 1:
```
MichaelClerx, line 414: What happens if you call this with a list of length 0? Think we should just return True if len() < 2. Easy to write a test for!

MauriceHendrix: I actually tries a little test:
```
>>> list_of_quantities = []
>>> list_of_quantities = iter(list_of_quantities)
>>> first = next(list_of_quantities, True)
>>> first
True
>>> return all(_is_equal(first, rest) for rest in list_of_quantities)
>>> all(_is_equal(first, rest) for rest in list_of_quantities)
True


>>> from cellmlmanip.model import Quantity
>>> from cellmlmanip.units import UnitStore
>>> list_of_quantities = [Quantity(2, UnitStore().get_unit('dimensionless'))]
>>> list_of_quantities = iter(list_of_quantities)
>>> first = next(list_of_quantities, True)
>>> first
_2
>>> all(_is_equal(first, rest) for rest in list_of_quantities)
True
```

cellmlmanip/printer.py
```diff
@@ -342,14 +334,11 @@ def _print_Piecewise(self, expr):
...
-                break
-            # Sympy filters these out:
-            # elif isinstance(c, BooleanFalse):
-            #    continue
-
-            # Add e-if-c-else-? statement
-            parts += self._print_ternary(c, e)
-            brackets += 1
+                # Sympy filters BooleanFalse out:
+            else:
```
jonc125, line 338: This is actually changing the logic: a piecewise with 3 cases where the middle one had `BooleanTrue` condition would now be handled differently.

tests/test_printer.py
```diff
@@ -222,6 +222,11 @@ def test_piecewise_expressions(self, p, x):
             evaluate=False)
         assert p.doprint(e) == '((0) if (x > 0) else (2))'
 
+        # First condition false, multiple true clauses
+        e = sp.Piecewise((6, False), (0, x < 5), (1, True), (2, True))
+        print(e)
+        assert p.doprint(e) == '((0) if (x < 5) else (1))'
+
```
jonc125, line 229: So you also want to test
```
        # Middle condition only true
        e = sp.Piecewise((0, x < 5), (1, True), (2, x > 7))
        assert p.doprint(e) == '((0) if (x < 5) else (1))'
```

tests/test_printer.py
```diff
@@ -222,6 +222,11 @@ def test_piecewise_expressions(self, p, x):
             evaluate=False)
         assert p.doprint(e) == '((0) if (x > 0) else (2))'
 
+        # First condition false, multiple true clauses
+        e = sp.Piecewise((6, False), (0, x < 5), (1, True), (2, True))
+        print(e)
```
jonc125, line 227: Cut the print?

cellmlmanip/units.py
MichaelClerx: Ah! I didn't see the next(.., True) part!

tests/test_units.py
```diff
@@ -255,9 +297,9 @@ def test_others(self):
         assert store.evaluate_units(sp.ceiling(12.6) * b) == store.get_unit('second')
         assert store.evaluate_units(sp.ceiling(_1)) == store.get_unit('kelvin')
 
-        # functions were all args should have same unit
-        # and this is the unit that will be returned
-        # fail cases
+    def test_func_args_same_units_fail_cases(self, store, a, b, x, y):
+        """ Tests functions were all args should have same unit and this is the unit returned, fail cases. """
```
jonc125, line 301:
```suggestion
        """ Tests functions where all args should have same unit and this is the unit returned, fail cases. """
```

tests/test_printer.py
MauriceHendrix: Yes that works fine!

It might seem to change the logic but that's not actually the case @jonc125 The reason is that sympy seems to filter out clauses after the True clause by the time they get to the printer so if I call
```
# Middle condition only true
e = sp.Piecewise((0, x < 5), (1, True), (2, x > 7))
p.doprint(e)'
```
then inside:
`_print_Piecewise` if I get  `expr`:
`Piecewise((0, x < 5), (1, True))`

Having said all that I don't really understand why adding the break makes it show up as a miss (or BrPart) in the coverage

jonc125: It'll be because the `break` is never *not* executed, so only one branch in the logic is taken. Similar to an `if` always being true, which is a coverage issue even with no `else` clause.

MauriceHendrix: Aah I see ok let me fix that

cellmlmanip/printer.py
MauriceHendrix: Not in practice as this would be filtered out (see below)

PR #312, opened 2020-07-04
Isort has stopped working

MauriceHendrix:
As you can see from the non-pinned python 3.6 test the isort config has stopped working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,3 @@ script:
   - isort --verbose --check-only --diff --recursive cellmlmanip
 after_success:
   - codecov
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
 script:
   - pytest --cov --cov-config=.coveragerc
   - flake8
-  - isort --verbose --check-only --diff --recursive cellmlmanip
+  - isort --verbose --check-only --diff cellmlmanip tests setup.py
 after_success:
   - codecov
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,4 @@ script:
   - isort --verbose --check-only --diff --recursive cellmlmanip
 after_success:
   - codecov
+
--- a/cellmlmanip/printer.py
+++ b/cellmlmanip/printer.py
@@ -324,6 +324,8 @@ class Printer(sympy.printing.printer.Printer):
         """
         from sympy.logic.boolalg import BooleanTrue
 
+
+
         # Assign NaN if no conditions hold
         # If a condition `True` is found, use its expression instead
         other = self._literal_names['nan']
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -25,3 +25,4 @@ script:
   - isort --verbose --check-only --diff --recursive cellmlmanip
```
MauriceHendrix:
```suggestion
  - isort --verbose --check-only --diff
```

MauriceHendrix: on codegen this seemed to fix the issue

MichaelClerx:
Seems to say 0 files changed now?

MauriceHendrix:
Yes the change is in the suggestion If I would have applied it I'm sure you guys would not be able to see why it's needed and would have rejected the proposed change.

MauriceHendrix:
> Github seems to say 0 files changed now in this PR?

Yes the change is in the suggestion If I would have applied it I'm sure you guys would not be able to see why it's needed and would have rejected the proposed change.

.travis.yml
```diff
@@ -25,3 +25,4 @@ script:
...
 after_success:
   - codecov
+
```
MauriceHendrix, line 28:
```suggestion
```

.travis.yml
MauriceHendrix, line 25:
```suggestion
  - isort --verbose --check-only --diff
```

MauriceHendrix: on codegen this seemed to fix the issue

jonc125:
So I would expect a PR like this to explain exactly what has stopped working and some indication of why, along with the actual fix. You could push one commit that demonstrates the broken behaviour, and then one that fixes it.

In this case the issue is that on Python 3.6+ the latest isort 5.0 release gets used, which as a major release contains backwards incompatibilities. See https://github.com/timothycrosley/isort/blob/develop/CHANGELOG.md#500-penny---july-4-2020 for details, in particular the note about the removal of the `--recursive` flag. It's not obvious from the short Travis result that there is a breakage since the test that uses the latest versions of all dependencies is an allowed failure (so we don't need to hold up minor PRs while fixing forward compatibility).

Now, on Python 3.5 we might want to keep the argument, since recursive processing is desired in general, but we can get away without it since we only have a single depth folder. This would break if we removed the `cellmlmanip` argument though. If we want import sorting in tests and setup.py we could use `isort --verbose --check-only --diff cellmlmanip tests setup.py` instead.

MauriceHendrix:
> So I would expect a PR like this to explain exactly what has stopped working and some indication of why, along with the actual fix. You could push one commit that demonstrates the broken behaviour, and then one that fixes it.

Sigh! :no_mouth:

Doesn't even work!

jonc125:
We can probably just remove the `not_skip` item from setup.cfg?

MauriceHendrix:
> We can probably just remove the `not_skip` item from setup.cfg?

that works

PR #313, opened 2020-10-27
Bump required PINT version to 0.16.0

MichaelClerx:
See https://github.com/ModellingWebLab/weblab-fc/issues/221

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,7 +9,7 @@ isodate==0.6.0            # via rdflib
 lxml==4.4.1
 mpmath==1.1.0             # via sympy
 networkx==2.3
-pint==0.16.0
+pint==0.9
 pyparsing==2.4.2          # via rdflib
 rdflib==4.2.2
 six==1.12.0               # via isodate
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,7 +9,7 @@ isodate==0.6.0            # via rdflib
 lxml==4.4.1
 mpmath==1.1.0             # via sympy
 networkx==2.3
-pint==0.9
+pint==0.16.0
 pyparsing==2.4.2          # via rdflib
 rdflib==4.2.2
 six==1.12.0               # via isodate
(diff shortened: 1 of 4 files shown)


MauriceHendrix:
odd that this decreases test coverage

MauriceHendrix (approved):
odd that this decreases test coverage

MauriceHendrix:
hhm @MichaelClerx  actually I just noticed with the requirement Pint>=0.16 on python3.5 it still installs Pint 0.9 anyway as that's the latest compatible version. It doesn't stop my tests from passing, but it might be an issue for yours.

MichaelClerx:
> hhm @MichaelClerx actually I just noticed with the requirement Pint>=0.16 on python3.5 it still installs Pint 0.9 anyway as that's the latest compatible version. It doesn't stop my tests from passing, but it might be an issue for yours.

Good find!
I think that was the pinned dependencies that Jonathan likes using. I've updated those to pint 0.16.0 too now, not sure if I've done that right 😅

MauriceHendrix:
I think Pint 0.16 just isn't compatible with Python 3.5

```
Could not find a version that satisfies the requirement pint>=0.16.0 (from cellmlmanip==0.2.0) (from versions: 0.1, 0.1.1, 0.1.2, 0.1.3, 0.2, 0.2.1, 0.3, 0.3.1, 0.3.2, 0.3.3, 0.4, 0.4.1, 0.4.2, 0.5, 0.5.1, 0.5.2, 0.6, 0.7, 0.7.1, 0.7.2, 0.8, 0.8.1, 0.9)
```

MichaelClerx:
In that case we might need to drop 3.5 support? If `pint` is no longer supporting it... @jonc125 ?

MichaelClerx:
Or we can just keep testing with 0.9 I guess, if we need 3.5 for chaste?

MauriceHendrix:
> Or we can just keep testing with 0.9 I guess, if we need 3.5 for chaste?

I would be slightly reluctant to pull python 3.5 at this stage, it's only 1 version below what e run on our server at Nottingham so not hard to imagine somebody else still running it on theirs.

MichaelClerx:
OK, it works with 0.9 as minimum required so let's just merge with that then?

PR #314, opened 2020-10-27
Support for secondary trig functions

MichaelClerx:
## Description
Cellmlmanip had partial support for secondary trig functions (sec, csc, cot, and their inverse and hyperbolic functions). This PR adds tests for the cellml-to-sympy transpiler and adds methods to the printer to write expressions to evaluate them (using python's `math` module, which doesn't have these expressions natively)

## Types of changes

## Documentation checklist

## Testing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,6 @@
+# Unreleased
+- Improved support for secondary trigonometric functions such as sec and acoth (PR#314).
+
 # Release 0.2.0
 - Speed of parsing CellML models has been improved by reorganising connection processing. (PR#303)
 - Removed `Model.connect_variables` as it is no longer needed by the parser. (PR#303)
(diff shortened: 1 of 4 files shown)


MauriceHendrix (approved):
I checked and it doesn't seem to interfere with chaste_codegen (doesn't change test results). 

Adding this to the chaste printer also doesn't add the test results, which would perhaps suggest that these don't appear in the test cellml files ( I didn't check that)

MichaelClerx:
Yeah I'd be very surprised if any of these show up in cardiac cell models. I can think of one with a `tanh` function, and one with an `acos`, but that's pretty much it as far as trig functions are concerned

MichaelClerx:
They are part of CellML though (for the stupidest reason: "it was easy to implement in C". I tried to reduce the number of trig functions a bit for CellML 2 but nobody else wanted too)

MauriceHendrix:
ok fair enough, so we should probably have them in chaste_codegen

PR #328, opened 2021-01-20
Moving to only testing on pull request

MauriceHendrix:
Let's run the builds only for pull requests rather than every push


MichaelClerx (approved):
If you're needing more inspiration/examples, Fergus has set up testing with actions for PINTS for us recently https://github.com/pints-team/pints

PR #334, opened 2021-03-17
Singularity fixes

MauriceHendrix:
## Types of changes

## Documentation checklist

- [ ] I have updated all documentation in the code where necessary.
- [ ] I have checked spelling in all (new) comments and documentation.

## Testing

### DIFF
--- a/cellmlmanip/parser.py
+++ b/cellmlmanip/parser.py
@@ -12,6 +12,7 @@ import sympy
 from lxml import etree
 
 from cellmlmanip.model import SYMPY_SYMBOL_DELIMITER, Model
+from cellmlmanip.singularity_fixes import fix_singularity_equations
 
 
 UNIT_PREFIXES = {
@@ -101,7 +102,7 @@ class Parser(object):
         # A dictionary mapping component names to _Component objects
         self.components = {}
 
-    def parse(self, unit_store=None):
+    def parse(self, unit_store=None, skip_singularity_fixes=False):
         """
         The main method that reads the XML file and extracts the relevant parts of the CellML model
         definition.
@@ -139,6 +140,20 @@ class Parser(object):
         # Canonicalise representation
         self.transform_constants()
 
+        if not skip_singularity_fixes:
+            try:
+                OXMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#'
+                PYCMLMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
+                # Get and convert V
+                V = self.model.get_variable_by_ontology_term((OXMETA, 'membrane_voltage'))
+                
+                tagged = set(self.model.get_variables_by_rdf((PYCMLMETA, 'modifiable-parameter'), 'yes', sort=False))
+                annotated = set(filter(lambda q: self.model.has_ontology_annotation(q, OXMETA), self.model.variables()))
+
+                modifiable_parameters = (tagged | annotated) - set(self.model.get_derived_quantities(sort=False) + [self.model.get_free_variable()]) - set(self.model.get_state_variables(sort=False))
+                fix_singularity_equations(self.model, V, modifiable_parameters, exp_function=SIMPLE_MATHML_TO_SYMPY_CLASSES['exp'])
+            except (KeyError, ValueError):
+                pass # V not tagged or no free variable
         return self.model
 
     @staticmethod
--- a/cellmlmanip/parser.py
+++ b/cellmlmanip/parser.py
@@ -46,6 +46,8 @@ class XmlNs(Enum):
     CMETA = 'http://www.cellml.org/metadata/1.0#'
     MATHML = 'http://www.w3.org/1998/Math/MathML'
     RDF = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'
+    OXMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#'
+    PYCMLMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
 
 
 def with_ns(ns_enum, name):
@@ -140,20 +142,27 @@ class Parser(object):
         # Canonicalise representation
         self.transform_constants()
 
+        # If enabled, equations are analysied for singularities in GHK equations
+        # and singularities found are fixed by inserting and appropriate piecewise
         if not skip_singularity_fixes:
             try:
-                OXMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#'
-                PYCMLMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
-                # Get and convert V
-                V = self.model.get_variable_by_ontology_term((OXMETA, 'membrane_voltage'))
-                
-                tagged = set(self.model.get_variables_by_rdf((PYCMLMETA, 'modifiable-parameter'), 'yes', sort=False))
-                annotated = set(filter(lambda q: self.model.has_ontology_annotation(q, OXMETA), self.model.variables()))
+                # Get V
+                V = self.model.get_variable_by_ontology_term((XmlNs.OXMETA.value, 'membrane_voltage'))
 
-                modifiable_parameters = (tagged | annotated) - set(self.model.get_derived_quantities(sort=False) + [self.model.get_free_variable()]) - set(self.model.get_state_variables(sort=False))
-                fix_singularity_equations(self.model, V, modifiable_parameters, exp_function=SIMPLE_MATHML_TO_SYMPY_CLASSES['exp'])
+                # Retreive variables for which to exclude analysys to make sure their equations remain.
+                # These are: modifiable-parameters, derived quantities,
+                # annotated variables (so that they will remain findable), time (the free variable), state variables
+                tagged = set(self.model.get_variables_by_rdf((XmlNs.PYCMLMETA.value, 'modifiable-parameter'), 'yes',
+                                                             sort=False))
+                annotated = set(filter(lambda q: self.model.has_ontology_annotation(q, XmlNs.OXMETA.value),
+                                       self.model.variables()))
+
+                excluded = (tagged | annotated) -\
+                    set(self.model.get_derived_quantities(sort=False) + [self.model.get_free_variable()]) -\
+                    set(self.model.get_state_variables(sort=False))
+                fix_singularity_equations(self.model, V, excluded, exp_function=SIMPLE_MATHML_TO_SYMPY_CLASSES['exp'])
             except (KeyError, ValueError):
-                pass # V not tagged or no free variable
+                pass  # V not tagged or no free variable
         return self.model
 
     @staticmethod
--- a/cellmlmanip/singularity_fixes.py
+++ b/cellmlmanip/singularity_fixes.py
@@ -17,9 +17,6 @@ from sympy import (
     exp
 )
 
-#from chaste_codegen._math_functions import exp_
-#from chaste_codegen._optimize import optimize_expr_for_c_output
-
 
 ONE = Quantity(1.0, 'dimensionless')
 
--- a/cellmlmanip/singularity_fixes.py
+++ b/cellmlmanip/singularity_fixes.py
@@ -1,0 +1,230 @@
+from math import isclose
+
+import networkx as nx
+from cellmlmanip.model import FLOAT_PRECISION, Quantity
+from sympy import (
+    Abs,
+    Add,
+    Eq,
+    Float,
+    I,
+    Mul,
+    Piecewise,
+    Pow,
+    Wild,
+    log,
+    solveset,
+    exp
+)
+
+#from chaste_codegen._math_functions import exp_
+#from chaste_codegen._optimize import optimize_expr_for_c_output
+
+
+ONE = Quantity(1.0, 'dimensionless')
+
+
+def _generate_piecewise(vs, ve, sp, ex, V):
+    """Generates a piecewsie for expression based on the singularity point sp and vmin (vs) and vmax (ve) """
+    def f(Vx, e):
+        return e.xreplace({V: Vx})
+
+    if vs is None:  # This shouldn't be apiecewise since we have no Vmin / Vmax
+        return ex
+
+    return Piecewise(*[(f(vs, ex) + ((V - vs) / (ve - vs)) * (f(ve, ex) - f(vs, ex)),
+                        Abs(V - sp) < Abs((ve - vs) / 2)), (ex, True)])
+
+
+def _get_U(expr, V, U_offset, exp_function):
+    '''Finds U in ghk equations these are of one of the the following forms where U is an expression over V:
+       - `U / (exp(U) - 1.0)`
+       - `U / (1.0 - exp(U))`
+       - `(exp(U) - 1.0) / U`
+       - `(1.0 - exp(U)) / U`
+       '''
+    Z = Wild('Z', real=True)
+    U_wildcard = Wild('U_wildcard', real=True, include=[V])
+    SP_wildcard = Wild('SP_wildcard', real=True)
+
+    def check_bottom_match(m):
+        return m is not None and U_wildcard in m and Z in m and Z != 0
+
+    def check_top_match(m, sp):
+        return m is not None and Z in m and Z != 0 and (sp == m[SP_wildcard]
+                                                        or (isinstance(sp, Float)
+                                                            and isinstance(m[SP_wildcard], Float)
+                                                            and isclose(m[SP_wildcard], sp)))
+
+    def float_dummies(expr):
+        '''Turns flaots back into Quantity dummies to be in line with the rest of the sympy equations'''
+        if expr is None:
+            return None
+        return expr.xreplace({f: Quantity(f, 'dimensionless') for f in expr.atoms(Float)})
+
+    # the denominator is all args where a **-1
+    numerator = tuple(a for a in expr.args if not isinstance(a, Pow) or a.args[1] != -1.0)
+    denominator = tuple(a.args[0] for a in expr.args if isinstance(a, Pow) and a.args[1] == -1.0)
+
+    # Not a devision or does not have exp
+    if len(denominator) == 0 or len(numerator) == 0 or not expr.has(exp_function):
+        return None, None, None
+
+    (vs, ve, sp) = None, None, None
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for num, denom in ((numerator, denominator), (denominator, numerator)):
+        found_on_top = False
+        for d in denom:  # Check arguments in denominator (or numerator)
+            if d.has(exp_function):
+                find_U = d.match(exp_function(U_wildcard) * -Z + 1.0)  # look for exp(U) * -Z + 1.0
+                if not check_bottom_match(find_U):
+                    find_U = d.match(exp_function(U_wildcard) * Z - 1.0)  # look for exp(U) * Z - 1.0
+                if check_bottom_match(find_U):
+                    # We found a match, since exp(U) * Z == exp(U + log(Z)) we can bring Z into the u expression
+                    u = (find_U[U_wildcard] + log(find_U[Z]))
+                    try:
+                        # Find the singularity point by solving u for V==0, excluding irratinal results
+                        sp = tuple(filter(lambda s: not s.has(I), solveset(u, V)))
+                        if sp:
+                            # we found a singularity point, now find vs,ve (or vmin, vmax)
+                            # by solving for U_offset either side of U
+                            find_v_low = solveset(u + U_offset, V)
+                            find_v_up = solveset(u - U_offset, V)
+                            find_v_low = tuple(find_v_low)
+                            find_v_up = tuple(find_v_up)
+                            if find_v_low and find_v_up:
+                                assert len(find_v_low) == len(find_v_up) == len(sp) == 1, \
+                                    'Expecting exactly 1 solution for singularity point'
+                                (vs, ve, sp) = (find_v_low[0], find_v_up[0], sp[0])
+                    except TypeError:
+                        pass  # Result could be 'ConditionSet' which is not iterable and not Real
+
+                if vs is not None:  # check top
+                    for n in num:  # Check arguments in numerator (or denominator)
+                        match = n.match(Z * V - Z * SP_wildcard)  # search for a multiple of V - sp
+                        found_on_top = check_top_match(match, sp)
+                        if not found_on_top:
+                            # search for a exp(multiple of V - sp)
+                            match = n.match(exp_function(Z * V - Z * SP_wildcard))
+                            found_on_top = check_top_match(match, sp)
+                            if not found_on_top:
+                                # A few equations don't play ball with the SP wildcard
+                                # We can find those by adding sp in the pattern directly
+                                match = n.match(Z * V - Z * sp)
+                                found_on_top = match is not None and Z in match and Z != 0
+                                if not found_on_top:
+                                    match = n.match(exp_function(Z * V - Z * sp))
+                                    found_on_top = match is not None and Z in match and Z != 0
+                        if found_on_top:  # We've found a match stop looking in the other numerator arguments
+                            break
+                    if found_on_top:  # found singularity, no need to try further
+                        break
+                    else:
+                        (vs, ve, sp) = None, None, None
+        if vs is not None and found_on_top:  # found singularity, no need to try further
+            break
+        else:
+            (vs, ve, sp) = None, None, None
+
+    # Put dummies back in and return the singularity point and range bundries
+    return (float_dummies(vs), float_dummies(ve), float_dummies(sp))
+
+
+def _new_expr_parts(expr, V, U_offset, exp_function):
+    """Removes suitable singularities and replaces it with a piecewise, returning (vs, ve, sp, has_singularity)"""
+
+    if isinstance(expr, Mul):  # 1 * A --> A (remove unneeded 1 *)
+        expr = Mul(*[a for a in expr.args if not str(a) in ('1.0', '1')])
+
+    # Turn Quantity dummies into numbers, to enable analysis
+    subs_dict = {d: d.evalf(FLOAT_PRECISION) for d in expr.atoms(Quantity)}
+    check_U_expr = expr.xreplace(subs_dict)
+
+    if not expr.has(exp_function):  # Expressions without exp don't have GHK equations
+        return (None, None, None, expr, False)
+
+    elif isinstance(expr, Add):  # A + B + ..
+        # The expression is an addition, find singularities in each argument
+        new_expr_parts = []
+        for a in expr.args:
+            new_expr_parts.append(_new_expr_parts(a, V, U_offset, exp_function))
+
+        # If all arguments have the same singularity point, return 1 singularity with the widest range
+        range = [item for (vs, ve, _, _, _) in new_expr_parts for item in (vs, ve)]
+        if len(new_expr_parts) > 1 and len(set([str(sp) for (_, _, sp, _, _) in new_expr_parts])) == 1 \
+                and all(isinstance(b, Quantity) for b in range):
+            sp = new_expr_parts[0][2]
+            vs, ve = min(range, key=lambda v: float(str(v))), max(range, key=lambda v: float(str(v)))
+            return (vs, ve, sp, expr, True)
+        else:  # Singularity points differ, so create each piecewise seperately and add them up
+            expr_parts = []
+            is_piecewise = False
+            for vs, ve, sp, ex, has_piecewise in new_expr_parts:
+                is_piecewise = is_piecewise or has_piecewise or vs is not None
+                expr_parts.append(_generate_piecewise(vs, ve, sp, ex, V))
+            return (None, None, None, Add(*expr_parts), is_piecewise)
+
+    elif isinstance(expr, Pow) and expr.args[1] == -1.0 and len(expr.args) == 2:  # 1/A
+        # Find singularities in A and adjust result to represent 1 / A
+        vs, ve, sp, ex, has_piecewise = _new_expr_parts(expr.args[0], V, U_offset, exp_function)
+        has_piecewise = has_piecewise or vs is not None
+        return (None, None, None, ONE / _generate_piecewise(vs, ve, sp, ex, V), has_piecewise)
+
+    elif isinstance(expr, Mul):  # A * B * ...
+        (vs, ve, sp) = _get_U(check_U_expr, V, U_offset, exp_function)  # Find the singularity point
+        if vs is not None:
+            return (vs, ve, sp, expr, True)
+        else:  # Couldn't find singularity, try the expression's arguments
+            expr_parts = []
+            is_piecewise = False
+            for sub_ex in expr.args:
+                vs, ve, sp, ex, has_piecewise = _new_expr_parts(sub_ex, V, U_offset, exp_function)
+                has_piecewise = is_piecewise = is_piecewise or has_piecewise or vs is not None
+                expr_parts.append(_generate_piecewise(vs, ve, sp, ex, V))
+            return (None, None, None, Mul(*expr_parts), is_piecewise)
+
+    else:  # Different type of equation e.g a number
+        return (None, None, None, expr, False)
+
+
+def new_expr(expr, V, U_offset=1e-7, exp_function=exp):
+    """Removes suitable singularities and replaces it with a piecewise.
+    :param: expr the expression to analyse ().
+    :param: V the volatge variable
+    :param: U_offset determins the offset either side of U for which the fix is used
+    :param: exp_function the function representing exp
+    :param: optimize whether or not to apply optimisations to the new expression
+    :return: expr with singularities fixes if appropriate
+    """
+    if not expr.has(exp_function):
+        return False, expr
+    (vs, ve, sp, ex, changed) = _new_expr_parts(expr, V, U_offset, exp_function)
+    ex = _generate_piecewise(vs, ve, sp, ex, V)
+    return changed or vs is not None, ex
+
+
+def fix_singularity_equations(model, V, modifiable_parameters, U_offset=1e-7, exp_function=exp):
+    """Finds singularities in the GHK equations in the model and replaces them with a piecewise.
+    :param: the cellmlmanip model the model to analyse.
+    :param: V the volatge variable
+    :param: modifiable_parameters the variables which are modifiable in the model,
+            their defining equations are excluded form the analysis
+    :param: U_offset determins the offset either side of U for which the fix is used
+    :param: exp_function the function representing exp
+    :param: optimize whether or not to apply optimisations to the new expression
+    """
+    unprocessed_eqs = {}
+    # iterate over sorted variables in the model
+    for variable in nx.lexicographical_topological_sort(model.graph, key=str):
+
+        # Get equation
+        eq = model.graph.nodes[variable]['equation']
+
+        # Skip variables that have no equation or equations defining parameters or where rhs is a Piecewise
+        if eq is not None and not isinstance(eq.rhs, Piecewise) and eq.lhs not in modifiable_parameters:
+            unprocessed_eqs[eq.lhs] = eq.rhs.xreplace(unprocessed_eqs)  # store partially evaluated version of the rhs
+            changed, new_ex = new_expr(unprocessed_eqs[eq.lhs], V, U_offset=U_offset, exp_function=exp_function)
+            if changed:  # update equation if the rhs has a singularity that can be fixed
+                model.remove_equation(eq)
+                model.add_equation(Eq(eq.lhs, new_ex))
+                unprocessed_eqs.pop(eq.lhs)
--- a/tests/test_singularity_fixes.py
+++ b/tests/test_singularity_fixes.py
@@ -1,0 +1,175 @@
+import pytest
+from sympy import (
+    Function,
+    Piecewise,
+    exp,
+    log,
+)
+
+from cellmlmanip.model import Quantity, Variable
+from cellmlmanip.parser import XmlNs
+from cellmlmanip.singularity_fixes import fix_singularity_equations, new_expr
+
+from . import shared
+
+
+def to_quant(val):
+    return Quantity(val, 'dimensionless')
+
+
+V = Variable(name='V', units='millivolt')
+Z = to_quant(6.2)
+U = to_quant(2) * V + to_quant(5.0)
+SP = (-(log(Z) + to_quant(5.0)) / to_quant(2))
+
+
+@pytest.fixture(scope='session')
+def expr1():
+    return (Z * V - Z * SP) / (exp(U) * -Z + to_quant(1.0))
+
+
+@pytest.fixture(scope='session')
+def expr2():
+    return (Z * V - Z * SP) / (exp(U) * Z - to_quant(1.0))
+
+
+class exp_(Function):
+
+    def _eval_is_real(self):
+        return self.args[0].is_real
+
+    def fdiff(self, argindex=1):
+        """
+        Returns the first derivative of this function.
+        """
+        assert argindex == 1
+        return self
+
+
+def test_no_singularity():
+    expr = (exp_(U) * -Z + to_quant(1.0)) / to_quant(25.0)
+    changed, new_ex = new_expr(expr, V)
+    assert str(new_expr(expr, V)) == str((False, expr))
+
+
+def test_no_singularity2():
+    expr = to_quant(1) / (to_quant(1) + exp_(to_quant(1.5669291338582676) - to_quant(0.07874015748031496) * V))
+    assert str(new_expr(expr, V)) == str((False, expr))
+
+
+def test_new_expr1(expr1):
+    print(str(new_expr(expr1, V)))
+    assert str(new_expr(expr1, V)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*(-(_-3.4122746960255230*_6.2 - '
+                                       '_6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_-3.4122746960255230*_2 + _5)) + (_'
+                                       '-3.4122745960255229*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_-3.4122'
+                                       '745960255229*_2 + _5)))/(_-3.4122745960255229 - _-3.4122746960255230) + (_-3.4'
+                                       '122746960255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_-3.41227469'
+                                       '60255230*_2 + _5)), Abs(_-3.4122746460255229 - _V) < Abs(_-3.4122745960255229/'
+                                       '2 - _-3.4122746960255230/2)), ((_6.2*_V - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.'
+                                       '2*exp(_2*_V + _5)), True)))')
+    assert str(new_expr(expr1, V, U_offset=1e-5)) == ('(True, Piecewise(((-_-3.4122796460255230 + _V)*(-(_-3.412279646'
+                                                      '0255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_-3.4'
+                                                      '122796460255230*_2 + _5)) + (_-3.4122696460255229*_6.2 - _6.2*('
+                                                      '-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_-3.4122696460255229*_2 + '
+                                                      '_5)))/(_-3.4122696460255229 - _-3.4122796460255230) + (_-3.4122'
+                                                      '796460255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp('
+                                                      '_-3.4122796460255230*_2 + _5)), Abs(_-3.4122746460255229 - _V) '
+                                                      '< Abs(_-3.4122696460255229/2 - _-3.4122796460255230/2)), ((_6.2'
+                                                      '*_V - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp(_2*_V + _5)), '
+                                                      'True)))')
+    assert str(new_expr(expr1, V, exp_function=exp_)) == ('(False, (_6.2*_V - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp'
+                                                          '(_2*_V + _5)))')
+    expr = expr1.replace(exp, exp_)
+    assert str(new_expr(expr, V, exp_function=exp_)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*(-(_-3.412274'
+                                                         '6960255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp_'
+                                                         '(_-3.4122746960255230*_2 + _5)) + (_-3.4122745960255229*_6.2'
+                                                         ' - _6.2*(-_5 - log(_6.2))/_2)/(_1 - _6.2*exp_(_-3.4122745960'
+                                                         '255229*_2 + _5)))/(_-3.4122745960255229 - _-3.41227469602552'
+                                                         '30) + (_-3.4122746960255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2'
+                                                         ')/(_1 - _6.2*exp_(_-3.4122746960255230*_2 + _5)), Abs(_-3.41'
+                                                         '22746460255229 - _V) < Abs(_-3.4122745960255229/2 - _-3.4122'
+                                                         '746960255230/2)), ((_6.2*_V - _6.2*(-_5 - log(_6.2))/_2)/(_1'
+                                                         ' - _6.2*exp_(_2*_V + _5)), True)))')
+
+
+def test_new_expr2(expr2):
+    assert str(new_expr(expr2, V)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*(-(_-3.4122746960255230*_6.2 - '
+                                       '_6.2*(-_5 - log(_6.2))/_2)/(-_1 + _6.2*exp(_-3.4122746960255230*_2 + _5)) + ('
+                                       '_-3.4122745960255229*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(-_1 + _6.2*exp(_-3.41'
+                                       '22745960255229*_2 + _5)))/(_-3.4122745960255229 - _-3.4122746960255230) + (_-3'
+                                       '.4122746960255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2)/(-_1 + _6.2*exp(_-3.41227'
+                                       '46960255230*_2 + _5)), Abs(_-3.4122746460255229 - _V) < Abs(_-3.41227459602552'
+                                       '29/2 - _-3.4122746960255230/2)), ((_6.2*_V - _6.2*(-_5 - log(_6.2))/_2)/(-_1 +'
+                                       ' _6.2*exp(_2*_V + _5)), True)))')
+
+
+def test_new_expr3(expr1):
+    assert str(new_expr(1 / expr1, V)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*((_1 - _6.2*exp(_-3.412274'
+                                           '5960255229*_2 + _5))/(_-3.4122745960255229*_6.2 - _6.2*(-_5 - log(_6.2))/_'
+                                           '2) - (_1 - _6.2*exp(_-3.4122746960255230*_2 + _5))/(_-3.4122746960255230*'
+                                           '_6.2 - _6.2*(-_5 - log(_6.2))/_2))/(_-3.4122745960255229 - _-3.41227469602'
+                                           '55230) + (_1 - _6.2*exp(_-3.4122746960255230*_2 + _5))/(_-3.4122746960255'
+                                           '230*_6.2 - _6.2*(-_5 - log(_6.2))/_2), Abs(_-3.4122746460255229 - _V) < Ab'
+                                           's(_-3.4122745960255229/2 - _-3.4122746960255230/2)), ((_1 - _6.2*exp(_2*_'
+                                           'V + _5))/(_6.2*_V - _6.2*(-_5 - log(_6.2))/_2), True)))')
+
+
+def test_new_expr4(expr2):
+    assert str(new_expr(1 / expr2, V)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*((-_1 + _6.2*exp(_-3.41227'
+                                           '45960255229*_2 + _5))/(_-3.4122745960255229*_6.2 - _6.2*(-_5 - log(_6.2))/'
+                                           '_2) - (-_1 + _6.2*exp(_-3.4122746960255230*_2 + _5))/(_-3.412274696025523'
+                                           '0*_6.2 - _6.2*(-_5 - log(_6.2))/_2))/(_-3.4122745960255229 - _-3.412274696'
+                                           '0255230) + (-_1 + _6.2*exp(_-3.4122746960255230*_2 + _5))/(_-3.4122746960'
+                                           '255230*_6.2 - _6.2*(-_5 - log(_6.2))/_2), Abs(_-3.4122746460255229 - _V) <'
+                                           ' Abs(_-3.4122745960255229/2 - _-3.4122746960255230/2)), ((-_1 + _6.2*exp('
+                                           '_2*_V + _5))/(_6.2*_V - _6.2*(-_5 - log(_6.2))/_2), True)))')
+
+
+def test_new_expr5(expr1):  # Try with numbers in place of Quantity Dummies
+    expr = expr1.xreplace({v: float(str(v)) for v in expr1.atoms(Quantity)})
+    print(new_expr(expr, V))
+    assert str(new_expr(expr, V)) == ('(True, Piecewise(((-_-3.41227469602552 + _V)*(-(6.2*_-3.41227469602552 + 21.156'
+                                      '1028053582)/(1.0 - 920.161586435975*exp(2.0*_-3.41227469602552)) + (6.2*_-3.412'
+                                      '27459602552 + 21.1561028053582)/(1.0 - 920.161586435975*exp(2.0*_-3.41227459602'
+                                      '552)))/(_-3.41227459602552 - _-3.41227469602552) + (6.2*_-3.41227469602552 + 21'
+                                      '.1561028053582)/(1.0 - 920.161586435975*exp(2.0*_-3.41227469602552)), Abs(_-3.4'
+                                      '1227464602552 - _V) < Abs(_-3.41227459602552/2 - _-3.41227469602552/2)), ((6.2*'
+                                      '_V + 21.1561028053582)/(1.0 - 920.161586435975*exp(2.0*_V)), True)))')
+
+
+def test_fix_singularity_equations():
+    # check piecewises in model without and with singularities fixed
+    model = shared.load_model('faber_rudy_2000', skip_singularity_fixes=True)
+    old_piecewises = tuple(eq.lhs for eq in model.equations if eq.rhs.has(Piecewise))
+    assert len(old_piecewises) == 14
+
+    V = model.get_variable_by_ontology_term((XmlNs.OXMETA.value, 'membrane_voltage'))
+    fix_singularity_equations(model, V, [])
+    new_piecewises = tuple(eq.lhs for eq in model.equations if eq.rhs.has(Piecewise))
+    assert len(new_piecewises) == 24
+
+    additional_piecewises = [v for v in new_piecewises if v not in old_piecewises]
+    assert str(additional_piecewises) == ('[_L_type_Ca_channel$I_CaCa, '
+                                          '_L_type_Ca_channel_d_gate$tau_d, '
+                                          '_fast_sodium_current_m_gate$alpha_m, '
+                                          '_L_type_Ca_channel$I_CaK, '
+                                          '_L_type_Ca_channel$I_CaNa, '
+                                          '_non_specific_calcium_activated_current$I_ns_K, '
+                                          '_non_specific_calcium_activated_current$I_ns_Na, '
+                                          '_rapid_delayed_rectifier_potassium_current_xr_gate$tau_xr, '
+                                          '_slow_delayed_rectifier_potassium_current_xs1_gate$tau_xs1, '
+                                          '_slow_delayed_rectifier_potassium_current_xs2_gate$tau_xs2]')
+
+
+def test_fix_singularity_equations2():
+    # check fixes via load_model
+    model = shared.load_model('bondarenko_szigeti_bett_kim_rasmusson_2004_apical', skip_singularity_fixes=True)
+    old_piecewises = tuple(str(eq.lhs) for eq in model.equations if eq.rhs.has(Piecewise))
+    assert len(old_piecewises) == 1
+
+    fixed_model = shared.load_model('bondarenko_szigeti_bett_kim_rasmusson_2004_apical', skip_singularity_fixes=False)
+    new_piecewises = tuple(eq.lhs for eq in fixed_model.equations if eq.rhs.has(Piecewise))
+    assert len(new_piecewises) == 2
+
+    additional_piecewises = [v for v in new_piecewises if str(v) not in old_piecewises]
+    assert str(additional_piecewises) == '[_slow_delayed_rectifier_potassium_current$alpha_n]'
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,5 +1,5 @@
 name: build
-on: [push]
+on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/cellmlmanip/main.py
+++ b/cellmlmanip/main.py
@@ -2,10 +2,10 @@
 from cellmlmanip.parser import Parser
 
 
-def load_model(path, unit_store=None):
+def load_model(path, unit_store=None, skip_singularity_fixes=False):
     """Parses a CellML file and returns a :class:`cellmlmanip.model.Model`.
 
     :param unit_store: Optional :class:`cellmlmanip.units.UnitStore` instance; if given the model will share the
         underlying registry so that conversions between model units and those from the provided store work.
     """
-    return Parser(path).parse(unit_store=unit_store)
+    return Parser(path).parse(unit_store=unit_store, skip_singularity_fixes=skip_singularity_fixes)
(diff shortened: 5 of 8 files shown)


RELEASE.md
```diff
@@ -1,3 +1,16 @@
+# (unreleased)
+- Added an automatic fix for removable singularities in GHK equations. This feature has been moved from chaste_codegen to allow more general use.
```
MichaelClerx, line 2: This looks like too much information for a changelog! Would suggest replacing this with e.g. "Added a method Model.fix_singularities()" and write this all up in the docstring there

cellmlmanip/parser.py
```diff
@@ -101,7 +104,7 @@ def __init__(self, filepath):
         # A dictionary mapping component names to _Component objects
         self.components = {}
 
-    def parse(self, unit_store=None):
+    def parse(self, unit_store=None, skip_singularity_fixes=False):
```
MichaelClerx, line 107: Can we make `fix_singularities` a method of `Model` that the user calls explicitly? I don't think any of the new code below has anything to do with parsing, so doesn't belong in the parser

MichaelClerx: That would give you a place to document the fixes as well, and explain that you need to oxmeta annotations for this to work etc.

cellmlmanip/parser.py
```diff
@@ -139,6 +142,27 @@ def parse(self, unit_store=None):
...
+        # and singularities found are fixed by inserting and appropriate piecewise
+        if not skip_singularity_fixes:
+            try:
+                # Get V
+                V = self.model.get_variable_by_ontology_term((XmlNs.OXMETA.value, 'membrane_voltage'))
+
+                # Retreive variables for which to exclude analysys to make sure their equations remain.
+                # These are: modifiable-parameters, derived quantities,
+                # annotated variables (so that they will remain findable), time (the free variable), state variables
+                tagged = set(self.model.get_variables_by_rdf((XmlNs.PYCMLMETA.value, 'modifiable-parameter'), 'yes',
```
MichaelClerx, line 155: Can we get rid of this and replace it by an oxmeta annotation?

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
+from math import isclose
```
MichaelClerx, line 1: Please add a docstring that explains what this module is for

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+
+import networkx as nx
+from sympy import (
```
MichaelClerx, line 4: Should this whole module be private?

Single cellmlmanip is a library, we need to carefully think about its API.

This code adds a public module `cellmlmanip.singularity_fixes` that has public classes `Abs`, `Add`, `FLOAT_PRECISION`, `ONE`, etc.
So if we're saying this module is public, then these should probably not be imported with a `from`, but just accessed via `sympy.Abs` etc.

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+)
+
+from cellmlmanip.model import FLOAT_PRECISION, Quantity
+
+
+ONE = Quantity(1.0, 'dimensionless')
+
+
+def _generate_piecewise(vs, ve, sp, ex, V):
+    """Generates a piecewsie for expression based on the singularity point sp and vmin (vs) and vmax (ve) """
```
MichaelClerx, line 26: piecewise, ``sp``, etc.

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    def f(Vx, e):
+        return e.xreplace({V: Vx})
+
+    if vs is None:  # This shouldn't be apiecewise since we have no Vmin / Vmax
+        return ex
+
+    return Piecewise(*[(f(vs, ex) + ((V - vs) / (ve - vs)) * (f(ve, ex) - f(vs, ex)),
```
MichaelClerx, line 33: Why does this use a function?

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+                        Abs(V - sp) < Abs((ve - vs) / 2)), (ex, True)])
+
+
+def _get_U(expr, V, U_offset, exp_function):
+    '''Finds U in ghk equations these are of one of the the following forms where U is an expression over V:
```
MichaelClerx, line 38: And why ``` instead of """ ? I'd prefer if we used """ consistently so that I can comment out blocks using ```

MichaelClerx: Can we get a better name than "get_U" ?

MichaelClerx: Also, these might appear in other equations so really would not mention GHK here explicitly.

MichaelClerx: I'm a bit puzzled why you return U without returning what kind of expression it was in? Does it not make a difference for the solution that you apply what the exact form of the equation is?

MichaelClerx: Needs some info about the input and output as well. Seems to return 3 expressions, rather than "U"

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+            break
+        else:
+            (vs, ve, sp) = None, None, None
+
+    # Put dummies back in and return the singularity point and range bundries
+    return (float_dummies(vs), float_dummies(ve), float_dummies(sp))
+
+
+def _new_expr_parts(expr, V, U_offset, exp_function):
+    """Removes suitable singularities and replaces it with a piecewise, returning (vs, ve, sp, has_singularity)"""
```
MichaelClerx, line 123: If it removes singularities why is it called "new_exp_parts" ??

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+                has_piecewise = is_piecewise = is_piecewise or has_piecewise or vs is not None
+                expr_parts.append(_generate_piecewise(vs, ve, sp, ex, V))
+            return (None, None, None, Mul(*expr_parts), is_piecewise)
+
+    else:  # Different type of equation e.g a number
+        return (None, None, None, expr, False)
+
+
+def new_expr(expr, V, U_offset=1e-7, exp_function=exp):
+    """Removes suitable singularities and replaces it with a piecewise.
```
MichaelClerx, line 180: What's a "suitable" singularity?

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    :param: expr the expression to analyse ().
+    :param: V the volatge variable
```
MichaelClerx, line 182: voltage

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    :param: U_offset determins the offset either side of U for which the fix is used
```
MichaelClerx, line 183: What's U?

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    :param: exp_function the function representing exp
+    :param: optimize whether or not to apply optimisations to the new expression
```
MichaelClerx, line 185: Doesn't seem to be there anymore

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    if not expr.has(exp_function):
+        return False, expr
+    (vs, ve, sp, ex, changed) = _new_expr_parts(expr, V, U_offset, exp_function)
+    ex = _generate_piecewise(vs, ve, sp, ex, V)
+    return changed or vs is not None, ex
+
+
+def fix_singularity_equations(model, V, modifiable_parameters, U_offset=1e-7, exp_function=exp):
+    """Finds singularities in the GHK equations in the model and replaces them with a piecewise.
```
MichaelClerx, line 196: Again, no need to mention GHK, it matches a particular form that _can_ appear inside GHK equations, but it doesn't check that or anything, nor is that something we'd want it to do

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    :param: the cellmlmanip model the model to analyse.
+    :param: V the volatge variable
```
MichaelClerx, line 198: Same comments as for `new_expr` apply.

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+                vs, ve, sp, ex, has_piecewise = _new_expr_parts(sub_ex, V, U_offset, exp_function)
+                has_piecewise = is_piecewise = is_piecewise or has_piecewise or vs is not None
+                expr_parts.append(_generate_piecewise(vs, ve, sp, ex, V))
+            return (None, None, None, Mul(*expr_parts), is_piecewise)
+
+    else:  # Different type of equation e.g a number
+        return (None, None, None, expr, False)
+
+
+def new_expr(expr, V, U_offset=1e-7, exp_function=exp):
```
MichaelClerx, line 179: Does this need to be public? Otherwise you could replace parts of the docstring with " see :meth:`fix_singularity_equations` for details "
(Or the other way around)

cellmlmanip/singularity_fixes.py
```diff
@@ -0,0 +1,219 @@
...
+    :return: expr with singularities fixes if appropriate
+    """
+    if not expr.has(exp_function):
+        return False, expr
+    (vs, ve, sp, ex, changed) = _new_expr_parts(expr, V, U_offset, exp_function)
+    ex = _generate_piecewise(vs, ve, sp, ex, V)
+    return changed or vs is not None, ex
+
+
+def fix_singularity_equations(model, V, modifiable_parameters, U_offset=1e-7, exp_function=exp):
```
MichaelClerx, line 195: Better names might be "fix_singularities_in_model" for this one and "fix_singularities_in_expression" for the method above?

MichaelClerx (changes requested):
Hi @MauriceHendrix !

Looks quite finished in places, and less code than I thought this would be!

But have a bunch of comments about docstrings, names, and code organisation that can be addressed before we look at the code itself :D

tests/test_singularity_fixes.py
```diff
@@ -0,0 +1,175 @@
...
+    assert str(new_expr(expr, V)) == str((False, expr))
+
+
+def test_no_singularity2():
+    expr = to_quant(1) / (to_quant(1) + exp_(to_quant(1.5669291338582676) - to_quant(0.07874015748031496) * V))
+    assert str(new_expr(expr, V)) == str((False, expr))
+
+
+def test_new_expr1(expr1):
+    print(str(new_expr(expr1, V)))
```
MichaelClerx, line 61: print statement

tests/test_singularity_fixes.py
```diff
@@ -0,0 +1,175 @@
...
+    assert str(new_expr(expr1, V)) == ('(True, Piecewise(((-_-3.4122746960255230 + _V)*(-(_-3.4122746960255230*_6.2 - '
```
MichaelClerx, line 62: Can you add a comment above each of these showing what the input and output equations are in a readable form?

cellmlmanip/parser.py
MauriceHendrix: I'd rather not as it would mean everybody would need to update thir cellml files.

MichaelClerx: I've never seen this NS before though. Is it a chaste thing?

MauriceHendrix: yup

MichaelClerx: Hmmm! I guess it doesn't rely on that namespace though, just knows to avoid variables tagged in a certain way, so it doesn't make the method less general?

MauriceHendrix: I could probably bring in these as an extra parameter I'll look at it tomorrow

MichaelClerx: I dunno! What do @mirams and @jonc125 think?

cellmlmanip/singularity_fixes.py
MauriceHendrix: aah! that's more historic, changed now.

cellmlmanip/singularity_fixes.py
MauriceHendrix: added reference to other function that explains it

cellmlmanip/singularity_fixes.py
MauriceHendrix: I renamed it but it's tricky. it's a recurive algorithm that eventually removes singularities, but also passes round intermediate reults

cellmlmanip/singularity_fixes.py
MauriceHendrix: documentatio updated

RELEASE.md
```diff
@@ -1,3 +1,9 @@
+# (unreleased)
+- added Model.remove_fixable_singularities to allow removing fixable singularities in the model's equations.
+
+
+  See for more details appendix B in: Johnstone, R. H. (2018). Uncertainty characterisation in action potential modelling for cardiac drug safety [PhD thesis]. University of Oxford. https://ora.ox.ac.uk/objects/uuid:0a28829c-828d-4641-bfb0-11193ef47195
```
MichaelClerx, line 5: Can this link go in the docs somewhere? If it stays here it should be made clear which part of the changelog it refers to and the link should maybe be in a proper markdown format. The `[PhD thesis]` bit could cause problems too?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+    if vs is None:  # This shouldn't be apiecewise since we have no Vmin / Vmax
+        return ex
+
+    f_vs = ex.xreplace({V: vs})
+    f_ve = ex.xreplace({V: ve})
+    return Piecewise((f_vs + ((V - vs) / (ve - vs)) * (f_ve - f_vs),
+                      Abs(V - sp) < Abs((ve - vs) / 2)), (ex, True))
+
+
+def _get_sinularity(expr, V, U_offset, exp_function):
```
MichaelClerx, line 41: missing a g --> singularity

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+
+import cellmlmanip
+from cellmlmanip.model import FLOAT_PRECISION
+
+
+ONE = cellmlmanip.Quantity(1.0, 'dimensionless')
+
+
+def _generate_piecewise(vs, ve, sp, ex, V):
+    """Generates a piecewise for expression based on the singularity point (sp) and vmin (vs) and vmax (ve) """
```
MichaelClerx, line 31: Something more like "Generates a new (piecewise) expression based on ``ex`` that returns a linear interpolation when variable ``V`` has a value between...

You could use a code block:
```
The returned expression between .. and .. is::

  f(vs) + (f(ve) - f(vs)) * (V - vs) / (ve - vs)

whenever::

  ...
  
``

Or even :math: if we're planning to allow latex in our docstrings at some point

File: cellmlmanip/_singularity_fixes.py [hunk 23]
```diff
@@ -0,0 +1,234 @@
...
+    if vs is None:  # This shouldn't be apiecewise since we have no Vmin / Vmax
+        return ex
+
+    f_vs = ex.xreplace({V: vs})
+    f_ve = ex.xreplace({V: ve})
+    return Piecewise((f_vs + ((V - vs) / (ve - vs)) * (f_ve - f_vs),
+                      Abs(V - sp) < Abs((ve - vs) / 2)), (ex, True))
```
MichaelClerx, line 38: Why when `|V - V_singularity|` is less than half the range from ve to vs ? I thought you'd use ve and vs as the cut-off points so that it's nice and continuous?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+)
+
+import cellmlmanip
+from cellmlmanip.model import FLOAT_PRECISION
+
+
+ONE = cellmlmanip.Quantity(1.0, 'dimensionless')
+
+
+def _generate_piecewise(vs, ve, sp, ex, V):
```
MichaelClerx, line 30: This order is quite weird: 3 values of V, then ex, then V again. Maybe ex, V, sp, vs, ve ?
And could they have nicer names, e.g. Vmin or Vlo and Vhi or something?
Also not sure why we need `sp` here (see below)

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+        return ex
+
+    f_vs = ex.xreplace({V: vs})
+    f_ve = ex.xreplace({V: ve})
+    return Piecewise((f_vs + ((V - vs) / (ve - vs)) * (f_ve - f_vs),
+                      Abs(V - sp) < Abs((ve - vs) / 2)), (ex, True))
+
+
+def _get_sinularity(expr, V, U_offset, exp_function):
+    """Finds U_offset and teh singularity either side and singularity U in ghk-like equations
```
MichaelClerx, line 42: Something like:

"Finds singularities in equations of the form::"   <-- two semicolons to start code block

Or

"Finds singularities in equations of the form:", and then "- \`\`lalala\`\`" with two \`\` (which is the syntax in restructured text. annoying, I know!)

And then below, "In addition to the singularity itself, this functions returns a lower and upper bounds such that..." and then explain whatever it is you do to determine these boundaries

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+        return m is not None and U_wildcard in m and Z in m and Z != 0
+
+    def check_top_match(m, sp):
+        return m is not None and Z in m and Z != 0 and (sp == m[SP_wildcard]
+                                                        or (isinstance(sp, Float)
+                                                            and isinstance(m[SP_wildcard], Float)
+                                                            and isclose(m[SP_wildcard], sp)))
+
+    def float_dummies(expr):
+        """Turns flaots back into Quantity dummies to be in line with the rest of the sympy equations"""
```
MichaelClerx, line 64: floats

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+            break
+        else:
+            (vs, ve, sp) = None, None, None
+
+    # Put dummies back in and return the singularity point and range bundries
+    return (float_dummies(vs), float_dummies(ve), float_dummies(sp))
+
+
+def _fixe_expr_parts(expr, V, U_offset, exp_function):
+    """Removes fixable singularities and replaces it with a piecewise
```
MichaelClerx, line 129: replaces them with piecewise expressions.

You'll need a newline after that too to make it render correctly I think

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+    :return: either (Vmin (vs), Vmax (ve), singularity point, expression, True)
+                    if we have identified a singularity needs to be made but it can't be done yet.
+                    For eample if 2 singularities with the same singularity point are added up,
+                    a single singularity is constructed instead.
+             or `(None, None, None, fixed expr, fixed_expr has piecewise)` otherwise
```
MichaelClerx, line 134: You need double \`\` in docstrings

MichaelClerx: Use `make html` in the `docs` subdirectory to check that this all compiles into nice looking html

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,234 @@
...
+
+    see :meth:fix_singularity_equations for more details "
```
MichaelClerx, line 136:
```
:meth:`fix_singularity_equations`
```

cellmlmanip/_singularity_fixes.py
MichaelClerx: Actually, since all this is hidden we won't be rendering it 😆 

Still need to carefully write these docstrings to explain what the code below is supposed to do though. Otherwise I can't review if it's doing that correctly!

MichaelClerx (commented):
Hi @MauriceHendrix, still some more things that aren't clear from the docstrings!

cellmlmanip/_singularity_fixes.py
MauriceHendrix: vs = solve (U-1e-7) for V
ve = vs = solve (U+1e-7) for V
sp = solve (U) for V
so ve - vs is the whole range 1e-7 either side of sp or if we substract sp:
1e-7 - -1e-7 = 2e-7 so that's twice what we want hence the half

MichaelClerx: so why not just write that into the equation instead of using 3 variables?

MichaelClerx: Oh wait, I don't understand what `solve` means in your example

MauriceHendrix: with solve (U+1e-7) for V I mean:
`solveset(u + 1e-7, V)`
that code gets me the value of V for which (U + 1e-7) == 0

MauriceHendrix: > so why not just write that into the equation instead of using 3 variables?

how would I write that?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,248 @@
...
+
+def _remove_singularities(expr, V, U_offset=1e-7, exp_function=exp):
+    """Removes suitable singularities and replaces it with a piecewise.
+    :param: expr the expression to analyse ().
+    :param: V the voltage variable
+    :param: U_offset determins the offset either side of U (see get_singularity) for which the fix is used
+    :param: exp_function the function representing exp
+    :return: (bool, expr) as follows: (expr has changed, expr with singularities fixes if appropriate)
+
+    see :meth:`fix_singularity_equations for more details `"
```
MichaelClerx, line 215: update method name

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,248 @@
...
+def _fixe_expr_parts(expr, V, U_offset, exp_function):
+    """Removes fixable singularities and replaces them with piecewise expressions.
+
+    :return: either (Vmin, Vmax, singularity point, expression, True)
+                    if we have identified a singularity needs to be made but it can't be done yet.
+                    For eample if 2 singularities with the same singularity point are added up,
+                    a single singularity is constructed instead.
+             or ``(None, None, None, fixed expr, fixed_expr has piecewise)`` otherwise
+
+    see :meth:`fix_singularity_equations for more details` "
```
MichaelClerx, line 150: update method name

MichaelClerx (changes requested):

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,248 @@
...
+       whenever sp is not Null, returns the original expression otherwise.
+
+       :param: The expression to turn into a piecewise
+       :param: V the voltage variable.
+       :param: the value of the singularity point.
+       :param: the value of the lower bound.
+       :param: the value of the upper bound.
+       :return: expr with a singularity exception around sp, if sp is not None else return the original expr
+    """
+    if sp is None:  # This shouldn't be apiecewise since we have no Vmin / Vmax
```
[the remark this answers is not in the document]
MauriceHendrix: lazieness! but I agree it's cleaner if it's checked outside so I changed it

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,248 @@
...
+)
+
+import cellmlmanip
+from cellmlmanip.model import FLOAT_PRECISION
+
+
+ONE = cellmlmanip.Quantity(1.0, 'dimensionless')
+
+
+def _generate_piecewise(expr, V, sp, Vmin, Vmax):
```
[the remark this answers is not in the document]
MauriceHendrix: it can't be internal as it's also used in _remove_singularities. I've tried to explain it better:
```
sp - the value of the singularity point: the value of V for whih U == 0
Vmin - the value of the lower bound: the value of V for which U - U_offset == 0
Vmax - the value of the upper bound: the value of V for which U + U_offset == 0
```
You can't just add 1e-7 to sp, this is something we did try but in some cases the range is too small and in other's it's too big (as in the interpolation doesn't match at the edges) when pickung the value of V for which U +/- 1e-7 ==0 we get a good (non-dimensionalised) range.
It would be possible to pass around U instead, but I decided not to as I didn't want to spread the solving of equations over different places in the code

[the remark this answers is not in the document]
MichaelClerx: Thanks!

tests/test_singularity_fixes.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def test_no_V(caplog):
+    model = shared.load_model('test_no_V_no_odes')
+    model.remove_fixable_singularities()
+    assert 'WARNING' in caplog.text
+    assert "Has no membrane_voltage tagged, cannot remove fixable singuarities." in caplog.text
+
+
+def test_fix_singularity_equations():
```
MichaelClerx, line 48: I think we should only have one of these "integration tests" on a static model file (ideally one that's in the repo already?)

The rest of the tests should be unit tests where you add a new equation to a model, of some specific form you want to test, and then you just use the method that checks that particular equation and see if it does it correctly.

E.g. in `_get_singularity` there's some complex pattern matching rules (check_bottom_match etc.) so you'd want at least one unit test here per pattern that we match, and then a bunch of tests with slight variations on that pattern that should not be matched, e.g. when you have

```
return m is not None and Z in m and Z != 0 and (
    sp == m[SP_wildcard] or (isinstance(sp, Float)and isinstance(m[SP_wildcard], Float) and isclose(m[SP_wildcard], sp)))
```
I'd expect a test where `m` is None, where Z isn't in m, where Z == 0 etc.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    if len(denominator) == 0 or len(numerator) == 0 or not expr.has(exp_function):
+        return None, None, None
+
+    (Vmin, Vmax, sp) = None, None, None
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for num, denom in ((numerator, denominator), (denominator, numerator)):
+        found_on_top = False
+        for d in denom:  # Check arguments in denominator (or numerator)
+            if d.has(exp_function):
+                find_U = d.match(exp_function(U_wildcard) * -Z + 1.0)  # look for exp(U) * -Z + 1.0
```
MichaelClerx, line 96: Are all numbers positive in sympy? Maybe add a comment about this?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    numerator = tuple(a for a in expr.args if not isinstance(a, Pow) or a.args[1] != -1.0)
+    denominator = tuple(a.args[0] for a in expr.args if isinstance(a, Pow) and a.args[1] == -1.0)
+
+    # Not a devision or does not have exp
+    if len(denominator) == 0 or len(numerator) == 0 or not expr.has(exp_function):
+        return None, None, None
+
+    (Vmin, Vmax, sp) = None, None, None
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for num, denom in ((numerator, denominator), (denominator, numerator)):
```
MichaelClerx, line 92: Maybe `num` and `denom` could be `a` and `b` or something? To avoid having `denom` being the numerator

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    - sp is the singularity point: the value of V for whih U == 0
+    - Vmin is the value of V for which U - U_offset == 0
+    - Vmax is the value of V for which U + U_offset == 0:
+    :return: (Vmin, Vmax, singularity point)
+    """
+    Z = Wild('Z', real=True)
+    U_wildcard = Wild('U_wildcard', real=True, include=[V])
+    SP_wildcard = Wild('SP_wildcard', real=True)
+
+    def check_bottom_match(m):
```
MichaelClerx, line 67: Is there a better name for these, e.g. "check_is_one_minus_exp" or something?
Maybe add a comment here saying what this check checks, and possibly something about the assumptions that are being made? (E.g. does sympy leave "exp() - 1" and "1 - exp()" in that form or does it do something weird with it?)

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+                                                            and isinstance(m[SP_wildcard], Float)
+                                                            and isclose(m[SP_wildcard], sp)))
+
+    def float_dummies(expr):
+        """Turns floats back into Quantity dummies to be in line with the rest of the sympy equations"""
+        if expr is None:
+            return None
+        return expr.xreplace({f: cellmlmanip.Quantity(f, 'dimensionless') for f in expr.atoms(Float)})
+
+    # the denominator is all args where a **-1
```
MichaelClerx, line 82: Add comment above saying what this does? E.g. "Find all fractions" ?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    numerator = tuple(a for a in expr.args if not isinstance(a, Pow) or a.args[1] != -1.0)
```
MichaelClerx, line 83: Is this guaranteed to find matching numerator/denominator pairs? Should there be a test to check that?
Maybe this would be better/safer as a single loop that finds fractions, and then make e.g. a list [(num, denom), (num, denom), ...] ?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for num, denom in ((numerator, denominator), (denominator, numerator)):
+        found_on_top = False
+        for d in denom:  # Check arguments in denominator (or numerator)
+            if d.has(exp_function):
+                find_U = d.match(exp_function(U_wildcard) * -Z + 1.0)  # look for exp(U) * -Z + 1.0
+                if not check_bottom_match(find_U):
+                    find_U = d.match(exp_function(U_wildcard) * Z - 1.0)  # look for exp(U) * Z - 1.0
+                if check_bottom_match(find_U):
+                    # We found a match, since exp(U) * Z == exp(U + log(Z)) we can bring Z into the u expression
```
MichaelClerx, line 100: can you add some more info about what exactly you've matched, before going into the fix? (so more stuff before the "since exp().." part please)

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+                            find_v_up = solveset(u - U_offset, V)
+                            find_v_low = tuple(find_v_low)
+                            find_v_up = tuple(find_v_up)
+                            assert find_v_low and find_v_up and len(find_v_low) == len(find_v_up) == len(sp) == 1, \
+                                'Expecting exactly 1 singularity point '
+                            (Vmin, Vmax, sp) = (find_v_low[0], find_v_up[0], sp[0])
+                    except TypeError:
+                        pass  # Result could be 'ConditionSet' which is not iterable and not Real
+
+                if Vmin is not None:  # check top
```
MichaelClerx, line 118: Is this just "if there wasn't a type error" ? Some way to write this bit slightly nicer? Maybe a `continue` in the "except TypeError" clause?
(And then can any of those lines throw a typeerror, or can we make it more specific?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    # Not a devision or does not have exp
+    if len(denominator) == 0 or len(numerator) == 0 or not expr.has(exp_function):
+        return None, None, None
+
+    (Vmin, Vmax, sp) = None, None, None
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for num, denom in ((numerator, denominator), (denominator, numerator)):
+        found_on_top = False
+        for d in denom:  # Check arguments in denominator (or numerator)
+            if d.has(exp_function):
```
MichaelClerx, line 95: You could lose a bit of indenting madness here by doing if not d.has(): continue ?
Feel free to ignore this comment :-)

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+                    for n in num:  # Check arguments in numerator (or denominator)
+                        match = n.match(Z * V - Z * SP_wildcard)  # search for a multiple of V - sp
+                        found_on_top = check_top_match(match, sp)
+                        if not found_on_top:
+                            # search for a exp(multiple of V - sp)
+                            match = n.match(exp_function(Z * V - Z * SP_wildcard))
+                            found_on_top = check_top_match(match, sp)
+                        if found_on_top:  # We've found a match stop looking in the other numerator arguments
+                            break
+                    if found_on_top:  # found singularity, no need to try further
```
MichaelClerx, line 128: What if there's two singularities in an equation?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+                        break
+                    else:
+                        (Vmin, Vmax, sp) = None, None, None
+        if Vmin is not None and found_on_top:  # found singularity, no need to try further
```
MichaelClerx, line 132: When is this bit required? I thought you unset VMin already in the lines above?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+            break
+        else:
+            (Vmin, Vmax, sp) = None, None, None
+
+    # Put dummies back in and return the singularity point and range bundries
```
MichaelClerx, line 137:
```suggestion
    # Put dummies back in and return the singularity point and range boundries
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+
+    if isinstance(expr, Mul):  # 1 * A --> A (remove unneeded 1 *)
+        expr = Mul(*[a for a in expr.args if not str(a) in ('1.0', '1')])
+
+    # Turn Quantity dummies into numbers, to enable analysis
+    subs_dict = {d: d.evalf(FLOAT_PRECISION) for d in expr.atoms(cellmlmanip.Quantity)}
+    check_U_expr = expr.xreplace(subs_dict)
+
+    if not expr.has(exp_function):  # Expressions without exp don't have GHK-like equations
```
MichaelClerx, line 161: More efficient to check this first, before removing 1* and dummy conversion?

MichaelClerx (changes requested):

cellmlmanip/__init__.py
```diff
@@ -1,3 +1,4 @@
 """Main module for loading, parsing and manipulating CellML models"""
 from ._config import __version__, __version_int__, version  # noqa
 from .main import load_model  # noqa
+from .model import Quantity  # noqa
```
jonc125, line 4: Is there a reason to add this in the main API?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    exp,
+    log,
+    solveset,
+)
+
+import cellmlmanip
+from cellmlmanip.model import FLOAT_PRECISION
+
+
+ONE = cellmlmanip.Quantity(1.0, 'dimensionless')
```
jonc125, line 26: Would this be better?
```suggestion
from .model import FLOAT_PRECISION, Quantity

ONE = Quantity(1.0, 'dimensionless')
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+
+def _generate_piecewise(expr, V, sp, Vmin, Vmax):
+    """
+    Generates a new (piecewise) expression based on expr with a linear interpolation when V is between Vmin and Vmax
+
+    The returned expression between Vmin and Vmax is::
+    f(vs) + (f(ve) - f(vs)) * (V - vs) / (ve - vs)
+
+    :param: The expression to turn into a piecewise
+    :param: V the voltage variable.
```
jonc125, line 37:
```suggestion
    :param expr: the expression to turn into a piecewise
    :param V: the voltage variable
```
etc?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+        return m is not None and Z in m and Z != 0 and (sp == m[SP_wildcard]
+                                                        or (isinstance(sp, Float)
+                                                            and isinstance(m[SP_wildcard], Float)
+                                                            and isclose(m[SP_wildcard], sp)))
+
+    def float_dummies(expr):
+        """Turns floats back into Quantity dummies to be in line with the rest of the sympy equations"""
+        if expr is None:
+            return None
+        return expr.xreplace({f: cellmlmanip.Quantity(f, 'dimensionless') for f in expr.atoms(Float)})
```
jonc125, line 80: Will units checking still be OK after this, i.e. these are really all pure numbers?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # the denominator is all args where a **-1
+    numerator = tuple(a for a in expr.args if not isinstance(a, Pow) or a.args[1] != -1.0)
+    denominator = tuple(a.args[0] for a in expr.args if isinstance(a, Pow) and a.args[1] == -1.0)
+
+    # Not a devision or does not have exp
```
jonc125, line 86:
```suggestion
    # Not a division or does not have exp
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+        for d in denom:  # Check arguments in denominator (or numerator)
+            if d.has(exp_function):
+                find_U = d.match(exp_function(U_wildcard) * -Z + 1.0)  # look for exp(U) * -Z + 1.0
+                if not check_bottom_match(find_U):
+                    find_U = d.match(exp_function(U_wildcard) * Z - 1.0)  # look for exp(U) * Z - 1.0
+                if check_bottom_match(find_U):
+                    # We found a match, since exp(U) * Z == exp(U + log(Z)) we can bring Z into the u expression
+                    u = (find_U[U_wildcard] + log(find_U[Z]))
+                    try:
+                        # Find the singularity point by solving u for V==0, excluding irratinal results
```
jonc125, line 103:
```suggestion
                        # Find the singularity point by solving u for V==0, excluding irrational results
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    def check_bottom_match(m):
+        return m is not None and U_wildcard in m and Z in m and Z != 0
+
+    def check_top_match(m, sp):
+        return m is not None and Z in m and Z != 0 and (sp == m[SP_wildcard]
+                                                        or (isinstance(sp, Float)
+                                                            and isinstance(m[SP_wildcard], Float)
+                                                            and isclose(m[SP_wildcard], sp)))
+
+    def float_dummies(expr):
```
jonc125, line 76: Could these functions-defined-in-a-function become private functions in the module instead?

cellmlmanip/model.py
```diff
@@ -11,6 +11,7 @@
 import rdflib
 import sympy
 
+import cellmlmanip
```
jonc125, line 14:
```suggestion
```

cellmlmanip/model.py
```diff
@@ -1041,6 +1043,53 @@ def _convert_variable_instance(self, original_variable, cf, units, direction, mo
...
+        See for more details appendix B in:
+        [Johnstone, R. H. (2018). Uncertainty characterisation in action potential modelling for cardiac drug safety.
+        University of Oxford.](https://ora.ox.ac.uk/objects/uuid:0a28829c-828d-4641-bfb0-11193ef47195)
+
+        :param exclude: set of variables which will not be substituted in the evaluation.
+
+        This ensures their defining equations will remain.
+        """
+        from ._singularity_fixes import remove_fixable_singularities
+        assert isinstance(exclude, set), 'eclude is expected to be a set'
```
jonc125, line 1069:
```suggestion
        assert isinstance(exclude, set), 'exclude is expected to be a set'
```
Though this should probably be a `raise TypeError` as it's complaining about user input.

cellmlmanip/model.py
```diff
@@ -1041,6 +1043,53 @@ def _convert_variable_instance(self, original_variable, cf, units, direction, mo
...
+
+        See for more details appendix B in:
+        [Johnstone, R. H. (2018). Uncertainty characterisation in action potential modelling for cardiac drug safety.
+        University of Oxford.](https://ora.ox.ac.uk/objects/uuid:0a28829c-828d-4641-bfb0-11193ef47195)
+
+        :param exclude: set of variables which will not be substituted in the evaluation.
+
+        This ensures their defining equations will remain.
+        """
+        from ._singularity_fixes import remove_fixable_singularities
```
jonc125, line 1068:
```suggestion
        from ._parser import SIMPLE_MATHML_TO_SYMPY_CLASSES
        from ._singularity_fixes import remove_fixable_singularities
```

cellmlmanip/model.py
```diff
@@ -1041,6 +1043,53 @@ def _convert_variable_instance(self, original_variable, cf, units, direction, mo
...
+            # annotated variables (so that they will remain findable), time (the free variable), state variables
+            annotated = set(filter(lambda q: self.has_ontology_annotation(q, OXMETA),
+                                   self.variables()))
+
+            excluded = (exclude | annotated) -\
+                set(self.get_derived_quantities(sort=False) + time) -\
+                set(self.get_state_variables(sort=False))
+
+            remove_fixable_singularities(self, V, excluded,
+                                         exp_function=cellmlmanip.parser.SIMPLE_MATHML_TO_SYMPY_CLASSES['exp'])
```
jonc125, line 1089:
```suggestion
                                         exp_function=SIMPLE_MATHML_TO_SYMPY_CLASSES['exp'])
```

cellmlmanip/model.py
```diff
@@ -1041,6 +1043,53 @@ def _convert_variable_instance(self, original_variable, cf, units, direction, mo
...
+        from ._singularity_fixes import remove_fixable_singularities
+        assert isinstance(exclude, set), 'eclude is expected to be a set'
+        try:
+            time = [self.get_free_variable()]
+        except ValueError:
+            time = []  # model has no free variable
+
+        try:
+            # Get V
+            V = self.get_variable_by_ontology_term((OXMETA, 'membrane_voltage'))
```
jonc125, line 1077: You could avoid the need for `cellmlmanip` to know about the oxmeta ontology by passing `V` in as an argument to this method?

jonc125 (commented):
I've not done an in-depth review, but had a quick skim through and it looks promising!

cellmlmanip/model.py
MauriceHendrix: That doesn't work when I change the exp_function, like I do in chaste_codegen (and from the tests)

cellmlmanip/model.py
MauriceHendrix: as per below: that doesn't work!

cellmlmanip/model.py
MauriceHendrix: It's needed (see below)

cellmlmanip/parser.py
mirams: Yeah, it's not really oxmeta in the sense that aims to annotate biology rather than model parameters/conversion settings etc. So this was just a way to preserve a variable/quantity when importing into Chaste in particular. It's a bit legacy so happy to leave as pycml name

cellmlmanip/model.py
MauriceHendrix: good idea

cellmlmanip/_singularity_fixes.py
MauriceHendrix: `float_dummies` yes but not the 2 working on the wildcards

cellmlmanip/_singularity_fixes.py
MauriceHendrix: yes they are. I checked it by adding this before the return and running the cellmlmanip & chaste-codegen test suite (no errors)
```
for f in expr.atoms(Float):
    float(f)
```

cellmlmanip/__init__.py
MauriceHendrix: I added it when I was struggeling with circular imports and forgot to take it out once I had figured out how to import it properly

cellmlmanip/_singularity_fixes.py
MauriceHendrix: good point! als by the way `.has(...)` is a lot quicker than `.atoms(..)`

cellmlmanip/_singularity_fixes.py
MauriceHendrix: Actually it makes things faster as you prevent re-checking existig Vmin etc from the previous loop (the initialisation is outside the main loop. It doesn't matter (if it didn't match last time it won't match now) but wastes time. However it's odd to read. I have added an initialisation just inside the `for d in denom` loop instead.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    :param: V the voltage variable.
+    :param: the value of the singularity point: the value of V for whih U == 0
+    :param: the value of the lower bound: the value of V for which U - U_offset == 0
+    :param: the value of the upper bound: the value of V for which U + U_offset == 0
+    :return: expr with a singularity exception around sp, if sp is not None else return the original expr
+    """
+    f_Vmin = expr.xreplace({V: Vmin})
+    f_Vmax = expr.xreplace({V: Vmax})
+    return Piecewise((f_Vmin + ((V - Vmin) / (Vmax - Vmin)) * (f_Vmax - f_Vmin),
+                      Abs(V - sp) < Abs((Vmax - Vmin) / 2)), (expr, True))
```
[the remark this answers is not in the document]
MauriceHendrix: why would there be a jump? It should be equivalent. 

The main reason for not going with `Vmin < V < Vmax` is that as far as I can see you can't create that in sympy (StrictLessThan takes exactly 2 parameters). Also it wouldn't render well at all in C++ `Vmin < V < Vmax`  === `(Vmin < V) < Vmax` where `(Vmin < V)` gets calculated and the 0/1 result is used to compare to Vmax

The proper alternative would be `(V < Vmax) & (Vmin < V)`
I think @mirams prefered the `fabs` version if I remember correctly?

cellmlmanip/_singularity_fixes.py
MauriceHendrix: Moved the solving into it's own function.
`_solve_tuple`

cellmlmanip/_singularity_fixes.py
MauriceHendrix: we can deal with `singularity + singularity + ...` and find both since both parts of the addition are treated seperately. 

But to be fair it is true we don't really deal with `singularity * singularity `. 

I ran some tests as follows (with num_singularities initialised to 0 before looping starts):

It counts and removes the simularity found from denom/numerator to prevent finding it when we swap numerator & denomenator.
```
                if found_on_top:  # found singularity, no need to try further
                    num_singularities +=1
                    assert num_singularities < 2
                    if denom == denominator:
                        denominator = [v for v in denominator if v != d]
                    else:
                        numerator = [v for v in numerator if v != d]
```

This is of course an ugly check. Let me know what you think.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,252 @@
...
+    - Vmin is the value of V for which U - U_offset == 0
+    - Vmax is the value of V for which U + U_offset == 0:
+    :return: (Vmin, Vmax, singularity point)
+    """
+    Z = Wild('Z', real=True)
+    U_wildcard = Wild('U_wildcard', real=True, include=[V])
+    SP_wildcard = Wild('SP_wildcard', real=True)
+
+    def check_bottom_match(m):
+        return m is not None and U_wildcard in m and Z in m and Z != 0
```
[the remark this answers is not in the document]
MauriceHendrix: uups should have been `m[Z] != 0` Z is the wildcard and the match m is dictionary

tests/test_singularity_fixes.py
MauriceHendrix: fair point I'll change it

cellmlmanip/_singularity_fixes.py
MauriceHendrix: This would because the numbers are in Quntity objects, otherwise you're right it wouldn't work. I'll add a comment to explain.

MauriceHendrix: The reason this does actually work, is that if Z matches a negative number, we later get log of a negative number and get irrational results which are skipped. However that's not very obvious so to make it more obvious I have added an explicit check to check the Z found > 0 (also negates some of the extra other checks I had to do for when solving resulted in an error)

cellmlmanip/_singularity_fixes.py
MauriceHendrix: Ok so I have now (yet to be pushed) made a fix and added a test, also fixed for the possibility of U==0 having multiple solutions

cellmlmanip/_singularity_fixes.py
MauriceHendrix: Fair point. I have updated it to include all negative powers (so that for example 1/V**2 which is V**(-2) would be found too. I've also added on the full numerator / denominator to allow the matching to match against the full expression rather than only a part

cellmlmanip/_singularity_fixes.py
MauriceHendrix: actually this isn't needed anymore

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    """
+    Generates a new (piecewise) expression based on expr with a linear interpolation when V is between Vmin and Vmax
+
+    The returned expression between Vmin and Vmax is::
+    f(Vmin) + (V - Vmin) / (Vmax - Vmin) * (f(Vmax) - f(Vmin))
+where ``f`` is ``expr``
+
+    :param expr: The expression to turn into a piecewise
+    :param V: The voltage variable.
+    :param sp: The value of the singularity point: the value of V for which U == 0.
```
MichaelClerx, line 46: Need to introduce U and U_offset here

cellmlmanip/_singularity_fixes.py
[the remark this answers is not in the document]
MichaelClerx: Yeah sure `(V < Vmax) and (Vmin < V)` is the sympy way to to write it. But what I'm saying is this is not equivalent to the abs version, because `Vmax != sp + (Vmax - Vmin) / 2` and `Vmin != sp - (Vmax - Vmin) / 2`.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    :return: expr with a singularity exception around sp, if sp is not None else return the original expr.
+    """
+    f_Vmin = expr.xreplace({V: Vmin})
+    f_Vmax = expr.xreplace({V: Vmax})
+    return Piecewise((f_Vmin + ((V - Vmin) / (Vmax - Vmin)) * (f_Vmax - f_Vmin),
+                      Abs(V - sp) < Abs((Vmax - Vmin) / 2)), (expr, True))
+
+
+def _float_dummies(expr):
+    """Turns floats back into Quantity dummies to be in line with the rest of the sympy equations"""
```
MichaelClerx, line 58:
```suggestion
    """Replaces any sympy ``Float`` objects in ``expr`` with (dimensionless) :class:`cellmlmanip.model.Quantity` objects."""
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    return expr.xreplace({f: Quantity(f, 'dimensionless') for f in expr.atoms(Float)})
+
+
+def _is_negative_power(expr):
+    try:
```
MichaelClerx, line 63:
```suggestion
    """
    Checks if ``expr`` is a sympy ``Pow`` with a negative exponent.
    
    If the exponent contains variables the sign of the exponent can't be determined, and ``False`` is returned.
    """
    try:
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+        return isinstance(expr, Pow) and bool(expr.args[1] < 0)
```
MichaelClerx, line 64: why `bool(... < ...)` ? Is the `<` operator overwritten to return something other than a boolean?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    except TypeError:  # if this is a power with variables still in it we can't determine if it's negative
```
MichaelClerx, line 65: We could fill in the values of the expression? If expression.evalf() < 0 ?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+        pass
+    return False
+
+
+def _solve_real(u, V):
+    """Gives the values of V for which u == 0 """
+    u = optimize(u, (_POW_OPT, ))  # make sure powers of ints are represented as ints
+    result = solveset(u, V, domain=S.Reals)
+    # The resul could be an intersection with Reals, if a custom exp function is used. We assume the result is real.
```
MichaelClerx, line 74: Don't quite understand this sentence

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    # Create "wildcards", that act as catch-all (*) when matching expressions
+    # https://docs.sympy.org/latest/modules/core.html#sympy.core.basic.Basic.match
+    P = Wild('P', real=True, exclude=[V])
+    Z = Wild('Z', real=True)
+    U_wildcard = Wild('U_wildcard', real=True, include=[V])
+    SP_wildcard = Wild('SP_wildcard', real=True)
+    singularities = []
+
+    def check_U_match(m, sp):
+        assert m is None or m[Z] != 0
```
MichaelClerx, line 104: Add comment or docstring saying what m and sp are

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+        return m is not None and \
+            (sp == m[SP_wildcard] or (isinstance(sp, Float) and
```
MichaelClerx, line 106: maybe a line break after `or (` just so the whole 2nd clause goes on one line?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+             isinstance(m[SP_wildcard], Float) and isclose(m[SP_wildcard], sp)))
+
+    # find all fractions and sperate numerator and denominator
```
MichaelClerx, line 109:
```suggestion
    # find all fractions and separate numerator and denominator
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    # the denominator is all args where a **-1
```
MichaelClerx, line 110:
```suggestion
    # The denominator is the set of arguments x inside a Pow(x, y) where y < 0
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    numerator = [a for a in expr.args if not _is_negative_power(a)]
+    denominator = [Pow(a.args[0], - a.args[1]) for a in expr.args if _is_negative_power(a)]
```
MichaelClerx, line 112: Is there some way to avoid testing is_negative_power(a) twice on all a? Maybe just a simple for loop instead of two list expressions? Not sure if that would be cheaper here? Just a thought :D

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # Not a division or does not have exp
+    if len(denominator) == 0 or len(numerator) == 0 or not expr.has(exp_function):
```
MichaelClerx, line 115:
```suggestion
    if len(denominator) == 0 or len(numerator) == 0:
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+        return m is not None and \
+            (sp == m[SP_wildcard] or (isinstance(sp, Float) and
+             isinstance(m[SP_wildcard], Float) and isclose(m[SP_wildcard], sp)))
+
+    # find all fractions and sperate numerator and denominator
+    # the denominator is all args where a **-1
+    numerator = [a for a in expr.args if not _is_negative_power(a)]
+    denominator = [Pow(a.args[0], - a.args[1]) for a in expr.args if _is_negative_power(a)]
+
+    # Not a division or does not have exp
```
MichaelClerx, line 114:
```suggestion
    # Second check: must be a fraction
```

cellmlmanip/_singularity_fixes.py
MichaelClerx: Instead of this check, you could just create the numerator and denominator (like below) and then check that both are functions of V?
Feels like that would be cheaper than starting the pattern matching before checking that?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # U might be on top, try numerator / denominator and denominator / numerator
+    for fraction_part_1, fraction_part_2 in ((numerator, denominator), (denominator, numerator)):
+        found_on_top = False
+        for fp2 in fraction_part_2:  # Check arguments in denominator (or numerator)
+            (Vmin, Vmax, sp) = None, None, None
+            if not fp2.has(exp_function):
+                continue
+            # look for exp(U) * -Z + 1.0
+            # Works as we later reuire Z to be positive
```
MichaelClerx, line 129:
```suggestion
            # Works as we later require Z to be positive
```

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    - Vmax is the value of V for which U + U_offset == 0:
+    :return: (Vmin, Vmax, sp)
+    """
+
+    # Create "wildcards", that act as catch-all (*) when matching expressions
+    # https://docs.sympy.org/latest/modules/core.html#sympy.core.basic.Basic.match
+    P = Wild('P', real=True, exclude=[V])
+    Z = Wild('Z', real=True)
+    U_wildcard = Wild('U_wildcard', real=True, include=[V])
+    SP_wildcard = Wild('SP_wildcard', real=True)
```
MichaelClerx, line 100: Is there a difference between the vars that have `_wildcard` in their names and the ones that don't?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+            (Vmin, Vmax, sp) = None, None, None
+            if not fp2.has(exp_function):
+                continue
+            # look for exp(U) * -Z + 1.0
+            # Works as we later reuire Z to be positive
+            find_U = fp2.match(exp_function(U_wildcard) * -Z + 1.0)
+            if not find_U:
+                find_U = fp2.match(exp_function(U_wildcard) * Z - 1.0)  # look for exp(U) * Z - 1.0
+
+            # Z should be positive, we replace free symbols by 1 to be able to evaluate the sign of Z
```
MichaelClerx, line 134: I don't understand this step. Why can't we evalf() Z to get its sign? Are free symbols just variables? And how do we know they are positive so that setting them to 1 is OK?

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    # The resul could be an intersection with Reals, if a custom exp function is used. We assume the result is real.
+    if isinstance(result, Intersection) and result.args[0] == S.Reals:
+        result = result.args[1]
+    return result
+
+
+def _get_singularity(expr, V, U_offset, exp_function):
+    """
+    Finds singularities in equations of the form:
+       - ``U / (exp(U) - 1.0)``
```
MichaelClerx, line 83: Further down we look for `exp(U) * -Z + 1.0`. Sounds like that should be mentioned here?

MichaelClerx (commented):
Thanks @MauriceHendrix ! I can look again tomorrow if you like, and then every day till we get it done 😄

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+    solveset,
+)
+from sympy.codegen.rewriting import ReplaceOptim, optimize
+from sympy.sets import Intersection
+
+from .model import FLOAT_PRECISION, Quantity
+
+
+# For P^n make sure n is passed as int if it is actually a whole number
+_POW_OPT = ReplaceOptim(lambda p: p.is_Pow and (isinstance(p.exp, Float) or isinstance(p.exp, float))
```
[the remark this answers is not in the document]
MauriceHendrix: You're eight and I tend to forget too.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,305 @@
...
+                        and float(p.exp).is_integer(),
+                        lambda p: Pow(p.base, int(float(p.exp))))
```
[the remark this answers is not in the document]
MauriceHendrix: Hhm good spot. I'm sure there was a rason but after removing and running tests and codegen & weblab tests it seems to not be needed, so I'm changing it to `float(p.exp)`

cellmlmanip/_singularity_fixes.py
MauriceHendrix: none, I just added _wildcard later as I was getting confused between singularity point value and the wildcard, I'll add _wildcard for all

cellmlmanip/_singularity_fixes.py
MauriceHendrix: actually `exp(U) * -Z + 1.0` 
is
`Z * ((1.0 - exp(U)) / U)`
so that's alradymentioned?

cellmlmanip/_singularity_fixes.py
MauriceHendrix: No but I've seen it not evaluate and just return ... < ... as an expression in some cases

cellmlmanip/_singularity_fixes.py
MauriceHendrix: When we get to this check all variables that have a value will already have been filled in. So if there are any remaining variables they will be either state variables, parameters or some sort of expression over those, so we don't know what the value is.

cellmlmanip/_singularity_fixes.py
MauriceHendrix: The result is usually a set. 
However if a custom function is used sympy doesn't know if the custom function is Real, in that case the result of solveset is a sympy Intersection between the actual result and Reals.

cellmlmanip/_singularity_fixes.py
MauriceHendrix: yes a single loop would be slighjtly faster

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,310 @@
...
+    Generates a new (piecewise) expression based on expr with a linear interpolation when V is between Vmin and Vmax
+
+    The returned expression between Vmin and Vmax is::
+    f(Vmin) + (V - Vmin) / (Vmax - Vmin) * (f(Vmax) - f(Vmin))
+where ``f`` is ``expr``
+
+    :param expr: The expression to turn into a piecewise
+    :param V: The voltage variable.
+    :param sp: The value of the singularity point: the value of V for which U == 0.
+    :param Vmin: The value of the lower bound: the value of V for which U - U_offset == 0.
```
MichaelClerx, line 47: Add note saying Vmin & Vmax aren't ordered...

cellmlmanip/_singularity_fixes.py
MauriceHendrix: I just ran some performance tests, when through all equations in 2 models and for the rhs tried the following 100 times and then averaged the timings:
x = eq.rhs.has(exp_function)
x = eq.rhs.atoms(exp_function)
x = V in eq.rhs.free_symbols
Average timings for the 2 models I tries:
5.717277526855469e-07 1.3425486428397042e-05 6.567750658307756e-06
1.73602785382952e-06 1.3979775565011159e-05 7.481302533830915e-06
So the .has is a pretty quick operation

MauriceHendrix: Having looked at the sorts of equations we get there's plenty which have V but no exp, but I've seen very few that do have exp not not V

cellmlmanip/_singularity_fixes.py
MauriceHendrix: free_symbols are just variables yes.

Z could contain a parameter. However it's a fair point this method only checks that there isn't a negative sign in this expression but the variables could represent negative numbers I'm not really sure how to check that. We could replace them by their initial values maybe? Not sure that's a good idea @MichaelClerx ?

MauriceHendrix: I've added some extra explanations explaining the limitation, my tests show it only happens in the synthetic tests, not in any real models.

cellmlmanip/_singularity_fixes.py
```diff
@@ -0,0 +1,357 @@
...
+
+    U here refers to the expression in the singularity witin the patter found
+    U_offset is the offset in U specified when finding Vmin / Vmax
+    see :meth:`_fix_expr_parts`
+    """
+    try:
+        if float(Vmax) < float(Vmin):
+            Vmin, Vmax = Vmax, Vmin
+        range_condition = And(Le(Vmin, V), Le(V, Vmax))
+    except TypeError:  # Vmax / Vmin are not a number, but contan variables
```
MichaelClerx, line 62:
```suggestion
    except TypeError:  # Vmax / Vmin are not a number, but contain variables
```

MichaelClerx (approved):

cellmlmanip/model.py
```diff
@@ -1041,6 +1043,34 @@ def _convert_variable_instance(self, original_variable, cf, units, direction, mo
...
+        - `U / (1.0 - exp(U))`
+        - `(exp(U) - 1.0) / U`
+        - `(1.0 - exp(U)) / U`
+        It replaces these with a piecewise 1e-7 either side of U==0 drawing a stright line in the region.
+        For example `(V + 5)/(exp(V + 5) - 1)` becomes
+        `((fabs(-V - 5.0000000000000000) < fabs(-4.9999999000000000 / 2 - -5.0000001000000000 / 2))
+        ? -0.494049243462503*V - 1.4702462167574 : ((5.0 + V) / (-1.0 + exp(5.0 + V))))`
+
+        See for more details appendix B in:
+        [Johnstone, R. H. (2018). Uncertainty characterisation in action potential modelling for cardiac drug safety.
```
MichaelClerx, line 1061: This isn't RST syntax!

MichaelClerx: is it?

cellmlmanip/_singularity_fixes.py
MauriceHendrix: yes I'm changing it to expr.args[1].evalf() just to be sure

cellmlmanip/model.py
MauriceHendrix: It's hard to find exactly what the right forwmat is but I think it's ok the way it's now (i'm making a small tweak).

PR #340, opened 2021-06-25
added in celsius as known base unit for parsing

### LINKED ISSUE #339: celsius is not defined in the unit registry
Tried to load up a model that OpenCOR is happy with, but cellmlmanip is complaining about it having units of celsius in there.
https://models.physiomeproject.org/workspace/5fd/@@file/e1c2a49185ddc44b0b55b1902dd08340a41e8f0b/  

Then tried to put in the unit definition in OpenCOR in terms of defining it in Kelvin with an offset, but then OpenCOR won't let me save that because celsius is already defined there!

I see they have been taken out of CellML2.0 but seem to appear in CellML 1.0 https://www.cellml.org/specifications/cellml_1.0 so we should probably include the definition to be nice?

MauriceHendrix:
## Description

adds celsius as a base unit used for parsing the model components

## Motivation and Context

fixes #339 

## Types of changes

## Documentation checklist

## Testing


mirams (commented):
I would think it is best described as a derived unit (just Kelvin minus 273.15) ? Or would an offset rather than scaling break things?

jonc125:
Leaving celsius out was a deliberate design decision rather than an oversight, precisely because it's not a base unit but derived with an offset, and almost all tools (and indeed the standard itself!) don't handle this properly if at all. It has been taken out of CellML 2.0 (as has all support for offsets) for this reason too.

mirams:
So we should manually re-write the model to Kelvin and remove any mention of celsius or what? This doesn't seem optimal if it is valid CellML1.0 that OpenCOR runs OK. I haven't looked for more models with celsius in, but I wouldn't be surprised to find some.

But I can see in general that we wouldn't want it as a base unit that can't be translated to Kelvin by WebLab etc. if requested, and it might be better to throw an error rather than mess that up silently.

MauriceHendrix:
> So we should manually re-write the model to Kelvin and remove any mention of celsius or what? This doesn't seem optimal if it is valid CellML1.0 that OpenCOR runs OK. I haven't looked for more models with celsius in, but I wouldn't be surprised to find some.
> 
> But I can see in general that we wouldn't want it as a base unit that can't be translated to Kelvin by WebLab etc. if requested, and it might be better to throw an error rather than mess that up silently.

Would it not just work if defined as an offset unit?

As it is if you try to define celsius in the cellml, cellmlmanip will throw this error:
```
ValueError: Cannot redefine CellML unit <celsius>.
```

MauriceHendrix:
> If there is less pain and suffering for all involved, I will just rewrite the equations and email a new version of the model to the author (it is actually a 2020 model). It is just used in Q10 calculations for one ion channel model
> 
> `(37 (Temp in celsius) - 20)/10`
> 
> I can make it
> 
> `(310 (Temp in Kelvin) - 293)/10`
> 
> and leave the maths unchanged. In this case the 20 is actually (wrongly) written as dimensionless so fixing it automatically might actually be dangerous!

yes but either way the current situation is a bug in my opinion what do you thing @jonc125

jonc125:
There is a bug in that the error is confusing. I would say cellmlmanip should throw an error saying that celsius is unsupported if models try using it.

MauriceHendrix:
> There is a bug in that the error is confusing. I would say cellmlmanip should throw an error saying that celsius is unsupported if models try using it.

really? So you wouldn't even allow redefining it?

jonc125:
You could argue on that point, but given the CellML spec says it should be built-in, redefining it is an error according to the spec...

MauriceHendrix:
> You could argue on that point, but given the CellML spec says it should be built-in, redefining it is an error according to the spec...

true

MauriceHendrix:
@jonc125 ok added a better error

mirams:
> @jonc125 ok added a better error

Have you pushed that Maurice, can't see any changes on here?

MauriceHendrix:
> > @jonc125 ok added a better error
> 
> Have you pushed that Maurice, can't see any changes on here?

uups I did give the command but it got rejected should be pushed now though.

RELEASE.md
```diff
@@ -1,5 +1,5 @@
 # Release 0.3.0
-- Added a method `Model.remove_fixable_singularities` to remove fixable singularities in the model's equations.
```
jonc125, line 2: Did you mean to remove the old release info?!

cellmlmanip/units.py
```diff
@@ -128,6 +131,9 @@ def __init__(self, store=None):
         # without the prefix to make them unique in the registry.
         self._known_units = set(_CELLML_UNITS)
 
+        # Names of units that are in the cellml spec but we currently do not support.
```
jonc125, line 134: I'd put this comment on the definition of _UNSUPPORTED_UNITS and not keep a local copy here - we never add to it after all.

cellmlmanip/units.py
```diff
@@ -149,8 +155,10 @@ def add_unit(self, name, expression):
         """
         if name in _CELLML_UNITS:
             raise ValueError('Cannot redefine CellML unit <%s>.' % name)
-        if name in self._known_units:
+        elif name in self._known_units:
             raise ValueError('Cannot redefine unit <%s>.' % name)
+        elif name in self._unsupported_units:
```
jonc125, line 160:
```suggestion
        elif name in _UNSUPPORTED_UNITS:
```

cellmlmanip/units.py
```diff
@@ -208,7 +216,9 @@ def get_unit(self, name):
         :returns: A :class:`Unit` object.
         :raises KeyError: If the unit is not defined in this unit store.
         """
-        if name not in self._known_units:
+        if name in self._unsupported_units:
```
jonc125, line 219:
```suggestion
        if name in _UNSUPPORTED_UNITS:
```

jonc125 (changes requested):
Looks good. Just a couple of minor changes needed I think.

RELEASE.md
MauriceHendrix: eeuh uups no!

jonc125 (approved):

Issue #8, opened 2017-11-16
Handle physical units


jonc125:
- [x] Need to parse Model level units
- [x] ~Need to parse Component level units, stored separately~
- [x] Need to resolve units to the correct level (component, model, predefined)
- [x] Need all numbers in equations to be quantities with units
- [x] Similarly all variables have units (although, see #21, these may not be known when the equation is first created, and need to be calculated when all model manipulation done) (issue #77 will deal with this)
- [x] Need to be able to calculate the units for an expression
- [x] Need to be able to automate units conversions (including adding explicit conversions in generated code) (issue #77 will deal with this)
    - [x] These are only required at connections between components
    - [x] The protocol can define conversion functions for cases where you're not just scaling within one dimension. These specify the source dimension (actually, arbitrary source units within that dimension), target dimension, and a unary function converting a value from source to target units. This gets to the correct dimension; scaling may then be needed to get the exact required units.
- [ ] Handle 'quirks' in the [CellML units system](https://www.cellml.org/specifications/cellml_1.1/#sec_units)
    - [x] The SI base units supported are: ampere, candela, kelvin, kilogram, meter, mole, second
    - [x] Modellers can define new 'base units' to create novel dimensions, unrelated to any other units
    - [x] There is an explicit 'dimensionless' unit for plain numbers
    - [x] There are some name aliases: metre/meter, litre/liter
    - [ ] pycml adds special 'boolean' units for the results of relational operators and the MathML symbols `true` and `false`. There could be other ways of handling this :)
    - [x] CellML does have an 'offset' field to support e.g. Celsius & Fahrenheit. But no models use this and the semantics are broken, so we'll say it's not allowed!
- [x] Handle typical 'quirks' in CellML models, e.g.
    - [x] `exp(V)` where `V` is in volts, not dimensionless (occasionally also in volt^2, millimolar, ...)
    - [x] `tanh(V)` where `V` is in volts, not dimensionless
    - Many models also have more serious errors, mostly dimensional mismatches in the operands to `eq`, `geq`, `gt`, `minus`, `plus`, `piecewise`. Currently we only warn about these when generating code, as otherwise we'd need to fix lots of CellML files!

MichaelClerx:
Ideas / initial libraries?

You mentioned https://pint.readthedocs.io/en/0.7.2/ at one stage?

jonc125:
Yes, we're using pint for several other projects, and it's working nicely there!

[...]

PR #354, opened 2022-09-27
fixing errors deling with offsets in unit definitions

### LINKED ISSUE #351: Maximum recursion depth exceeded
```
<!-- CellML 1.0, 5.2.7: The exponent of dimensionless doesn't matter -->
<model name="unit_conversion_prefix"
  <units name="hyper_dimensionless">
    <unit units="dimensionless" exponent="12" />
  </units>
  <component name="A">
    <variable name="x" units="dimensionless" initial_value="3" public_interface="out" />
  </component>
  <component name="B">
    <variable name="y" units="hyper_dimensionless" public_interface="in" />
  </component>
  <connection>
    <map_components component_1="A" component_2="B" />
    <map_variables variable_1="x" variable_2="y" />
  </connection>
</model>
```

MauriceHendrix:
## Description

Fixes errors dealing with:
- offsets in unit definitions (warning if non-0 and not dimensionless)
- offsets in demenionless units (error)
- demenionless units with an exponent
- demenionless units with a multiplier

## Motivation and Context

fixes #351

## Types of changes

## Documentation checklist

## Testing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,6 +2,9 @@
 - Fixed a bug in the parser where equations in a piecewise containing a boolean caused parsing errors.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/350
 
+- Fixes errors dealing with demenionless units which have an offset, multiplier or exponent and implements offset units more generally
+  see https://github.com/ModellingWebLab/cellmlmanip/issues/351
+
 # Release 0.3.4
 - Updated how substitution of functions that were changed in the parser are handled during analysis for fixing singularities, in order to make sure it workes with sympy 1.10
 - Dropped support for python 3.5 as it is end of life.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 - Fixed a bug in the parser where equations in a piecewise containing a boolean caused parsing errors.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/350
 
-- Fixes errors dealing with demenionless units which have an offset, multiplier or exponent and implements offset units more generally
+- Fixes errors dealing with dimensionless units which have an offset, multiplier or exponent and added a warbing for other offset units to indicate that these are not supported.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/351
 
 # Release 0.3.4
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 - Fixed a bug in the parser where equations in a piecewise containing a boolean caused parsing errors.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/350
 
-- Fixes errors dealing with dimensionless units which have a multiplier or exponent and added an error for offset units (where the offset is not 0) as those are not supported.
+- Fixed errors dealing with dimensionless units which have a multiplier or exponent, and added an error for offset units (where the offset is not 0) as those are not supported.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/351
 
 # Release 0.3.4
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 - Fixed a bug in the parser where equations in a piecewise containing a boolean caused parsing errors.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/350
 
-- Fixes errors dealing with dimensionless units which have an offset, multiplier or exponent and added a warbing for other offset units to indicate that these are not supported.
+- Fixes errors dealing with dimensionless units which have a multiplier or exponent and added an error for offset units (where the offset is not 0) as those are not supported.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/351
 
 # Release 0.3.4
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -293,7 +293,7 @@ class UnitStore(object):
         #  Trying to convert dimensionless gives an error but we can convert to it
         if quantity.units == self._registry.dimensionless:
             quantity, unit = 1 * unit, quantity.units
-            return 1 / quantity.to(unit)
+            return quantity.magnitude / quantity.to(unit)
         return quantity.to(unit)
 
     def add_conversion_rule(self, from_unit, to_unit, rule):
--- a/cellmlmanip/units.py
+++ b/cellmlmanip/units.py
@@ -168,15 +168,14 @@ class UnitStore(object):
         # To test if this is a dimensionless unit, parse the string as a Quantity and check if it's dimensionless
         quantity = self._registry.parse_expression(expression)
         if quantity.units == self._registry.dimensionless:
-            #  offsets are not supported for dimensionless
+            # offsets are not supported for dimensionless
             if offset is not None:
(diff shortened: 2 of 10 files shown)


MichaelClerx:
The offset bit won't work. Offsets are supposed to change the unit, which `pint` doesn't support, so we don't implement it in cellmlmanip (see here: https://github.com/ModellingWebLab/cellmlmanip/issues/8)

The presence of an offset that isn't 0 should be a warning or an error for all units (dimensionless or not)

MauriceHendrix:
now throwing error for dimentionless + offset other non-0 offsets

MichaelClerx:
```
[michael@fedora cellmlmanip]$ grep "upported" ./cellmlmanip/ -Ir
./cellmlmanip/units.py:# The full list of supported CellML units
./cellmlmanip/units.py:            raise ValueError('Unit <%s> is not currently supported by cellmlmanip.' % name)
./cellmlmanip/units.py:            raise KeyError('Unit <' + str(name) + '> is not currently supported by cellmlmanip.')
./cellmlmanip/units.py:        :raises UnexpectedMathUnitsError: if math is not supported
./cellmlmanip/units.py:        self.message = message or 'The math used by this expression is not supported.'
./cellmlmanip/parser.py:                    'Defining units inside components is not supported (found in component ' + name + ').')
./cellmlmanip/parser.py:                    'Reactions are not supported (found in component ' + name + ').')
./cellmlmanip/parser.py:            raise ValueError('Invalid or unsupported CellML file. ' + msg)
./cellmlmanip/model.py:                    'Non-local annotations are not supported.')
./cellmlmanip/model.py:        Note that only cmeta ids on variables or the model itself are checked and supported.
./cellmlmanip/model.py:                raise ValueError('Only first order derivatives wrt a single variable are supported')
./cellmlmanip/printer.py:            'Unsupported expression type (' + str(type(expr)) + '): '
./cellmlmanip/printer.py:        raise ValueError('Unsupported function: ' + str(name))
./cellmlmanip/printer.py:            raise ValueError('Unsupported relational: "' + str(op) + '".')
```
looks like we just use `ValueError` everywhere else where something isn't supported. Should probably be the same here? Also not sure why we need special logic for dimensionless-with-offset ? Any why pass the offset all the way through instead of raising it in the parser?

MauriceHendrix:
aah sorry I thought you wanted to make a distinction between dimentionless + offset and non-dimensionless + non-0 offset :) I've changed it to throw ValueError for non-0 offsets and indeed no need to carry the offset.

RELEASE.md
```diff
@@ -2,6 +2,9 @@
 - Fixed a bug in the parser where equations in a piecewise containing a boolean caused parsing errors.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/350
 
+- Fixes errors dealing with dimensionless units which have a multiplier or exponent and added an error for offset units (where the offset is not 0) as those are not supported.
```
MichaelClerx, line 5:
```suggestion
- Fixed errors dealing with dimensionless units which have a multiplier or exponent, and added an error for offset units (where the offset is not 0) as those are not supported.
```

tests/cellml_files/test_offset.cellml
```diff
@@ -0,0 +1,19 @@
...
+  <component name="A">
+    <variable name="x" units="meter" initial_value="3" public_interface="out" />
+  </component>
+  <component name="B">
+    <variable name="y" units="offsetmeter" public_interface="in" />
+  </component>
+  <connection>
+    <map_components component_1="A" component_2="B" />
+    <map_variables variable_1="x" variable_2="y" />
+  </connection>
```
MichaelClerx, line 18:
```suggestion
```

this bit's not needed for the test

tests/cellml_files/test_offset_0.cellml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- CellML Test Suite. https://github.com/MichaelClerx/cellml-validation -->
+<!-- CellML 1.0, 5.2.7: Dimensionless units can have an offset -->
```
MichaelClerx, line 3:
```suggestion
cellmlmaniptest: offset of 0 is ignored: does not affect parsing or unit equivalence when connections are made
```

MichaelClerx (changes requested):

tests/test_parser.py
```diff
@@ -331,3 +331,37 @@ def test_piecewise_booleans_error(self):
...
+        assert cf1 == cf2 == 1
+
+    def test_dimensionless_multiplier(self):
+        model = load_model('dimensionless_multiplier.cellml')
+        assert sorted(map(str, model.variables())) == ['A$x', 'B$y']
+        dimensionless = model.units.get_unit('dimensionless')
+        halves = model.units.get_unit('halves')
+        cf1 = model.units.convert(1 * dimensionless, halves)
+        cf2 = model.units.convert(1 * halves, dimensionless)
+        assert cf1 == 1 / cf2, str([cf1, cf2])
```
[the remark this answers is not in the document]
MichaelClerx: Could even think of making a copy of the test but with a unit, and then checking that we do the same as pint does?

So parse
```
<model name="unit_conversion_dimensionless_multiplier_1"
  <units name="half_meter">
    <unit units="meter" multiplier="0.5" />
  </units>
</model>
```
and then check that 1 meter gets converted to 2 half_meters

cellmlmanip/units.py
```diff
@@ -290,6 +290,10 @@ def convert(self, quantity, unit):
         """
         assert isinstance(quantity, self._registry.Quantity)
         assert isinstance(unit, self._registry.Unit)
+        #  Trying to convert dimensionless gives an error but we can convert to it
+        if quantity.units == self._registry.dimensionless:
+            quantity, unit = 1 * unit, quantity.units
+            return 1 / quantity.to(unit)
```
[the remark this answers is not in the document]
MauriceHendrix: Indeed it looks like a pint error, you can convert one way, but not the other way.
actually I think it should be `quantity.magnitude / quantity.to(unit)` rather than `1 / quantity.to(unit)`
so that 2 meters become 4 half meters en 3 half meters become 1.5 meter i.e.
```
    def test_dimensionless_multiplier2(self):
        model = load_model('dimensionless_multiplier2.cellml')
        meter = model.units.get_unit('meter')
        half_meter = model.units.get_unit('half_meter')
        cf1 = model.units.convert(2 * meter, half_meter).magnitude
        cf2 = model.units.convert(3 * half_meter, meter).magnitude
        assert cf1 == 4.0
        assert cf2 == 1.5
```

cellmlmanip/units.py
```diff
@@ -290,6 +290,10 @@ def convert(self, quantity, unit):
         """
         assert isinstance(quantity, self._registry.Quantity)
         assert isinstance(unit, self._registry.Unit)
+        #  Trying to convert dimensionless gives an error but we can convert to it
```
MichaelClerx, line 293:
```suggestion
        #  Trying to convert FROM dimensionless gives an error but we can convert TO it
```

MichaelClerx (approved):
Just one question left

cellmlmanip/units.py
MauriceHendrix: yes that's it exactly!

PR #359, opened 2022-09-29
improve units in components errors

### LINKED ISSUE #358: Error if component contains units
We're not supporting this (it will be gone in CellML 2), so should raise error if found

```
<!-- CellML 1.0, 3.4.3.3: A variable can have a local component unit -->
<model name="variable_units_component"
  <component name="A">
    <units name="oranges">
      <unit units="volt" />
    </units>
    <variable name="a" units="oranges" />
  </component>
</model>
```

MauriceHendrix:
## Description

Improved error reporting: When a unit is defined inside a component and used it now throws a ValueError indicating units in components are not supported. Previously a confusing KeyError was thrown, which suggested the unit was not found.
 
## Motivation and Context

Fixes #358 

## Types of changes

## Documentation checklist

## Testing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,7 @@
 - Added an error for duplicate component names.
 - Fixed errors dealing with dimensionless units which have a multiplier or exponent, and added an error for offset units (where the offset is not 0) as those are not supported.
   see https://github.com/ModellingWebLab/cellmlmanip/issues/351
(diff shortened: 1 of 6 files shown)


MichaelClerx (changes requested):
This seems an odd place to handle it! Can we just do it when we start looking at components, before variables or anything? Could even be a separate check right at the start (search for units in components with xpath or something?)

MauriceHendrix:
couldn't put it in components as you'd already have got a KeyError from the unit definitions by the time you get there, but pulling it out and using xpath is a good shout & now implemented

PR #383, opened 2026-07-15
Support sympy 1.13 and 1.14

### LINKED ISSUE #376: Support sympy 1.13
## Summary

Add support for sympy 1.13

Release notes: https://github.com/sympy/sympy/wiki/release-notes-for-1.13.0

kwabenantim:
## Description

Fixes #376

## Motivation and Context

Support latest sympy versions

## Types of changes

## Documentation checklist

## Testing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 - Migrated to Pint >= 0.24 (the `ScaleConverter` and `UnitDefinition` classes moved to `pint.facets.plain` and changed argument signatures). This reverses the `Pint < 0.20` pin introduced in 0.3.5 and restores compatibility with numpy 2.x, which older Pint versions could not import.
 - Added support for Python 3.14.
 - Dropped support for Python 3.9 as Pint >= 0.24 requires Python >= 3.10.
-- Added support for sympy 1.13 and 1.14.
+- Added support for sympy 1.13 and 1.14 (minimum supported sympy version is now 1.10).
 
 # Release 0.3.7
 - Added support for Python 3.12 and 3.13.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,6 +2,7 @@
 - Migrated to Pint >= 0.24 (the `ScaleConverter` and `UnitDefinition` classes moved to `pint.facets.plain` and changed argument signatures). This reverses the `Pint < 0.20` pin introduced in 0.3.5 and restores compatibility with numpy 2.x, which older Pint versions could not import.
 - Added support for Python 3.14.
 - Dropped support for Python 3.9 as Pint >= 0.24 requires Python >= 3.10.
+- Added support for sympy 1.13 and 1.14.
 
 # Release 0.3.7
 - Added support for Python 3.12 and 3.13.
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -27,9 +27,9 @@ classifiers = [
 dependencies = [
     "lxml>=4.7",
     "networkx>=2.1",
-    "Pint>=0.24, <0.26",
+    "Pint>=0.24, <0.26",  # Occasional breaking changes in minor versions
     "rdflib>=4",
-    "sympy>=1.4, <1.13",
+    "sympy>=1.10, <1.15",  # Occasional breaking changes in minor versions
 ]
```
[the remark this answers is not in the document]
kwabenantim: This is intentional. Sympy 1.10 is the oldest version on Python 3.10, which is now the lowest supported Python.

RELEASE.md
```diff
@@ -2,6 +2,7 @@
 - Migrated to Pint >= 0.24 (the `ScaleConverter` and `UnitDefinition` classes moved to `pint.facets.plain` and changed argument signatures). This reverses the `Pint < 0.20` pin introduced in 0.3.5 and restores compatibility with numpy 2.x, which older Pint versions could not import.
 - Added support for Python 3.14.
 - Dropped support for Python 3.9 as Pint >= 0.24 requires Python >= 3.10.
+- Added support for sympy 1.13 and 1.14.
```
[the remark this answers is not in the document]
kwabenantim: Fixed in 1d2e04d

cellmlmanip/_singularity_fixes.py
```diff
@@ -293,7 +293,7 @@ def _fix_expr_parts(expr, V, U_offset, exp_function):
                 expr_parts.append(_generate_piecewise(ex, V, sp, Vmin, Vmax) if sp is not None else ex)
             return (None, None, None, Add(*expr_parts), is_piecewise)
 
-    elif isinstance(expr, Pow) and expr.args[1] == -1.0 and len(expr.args) == 2:  # 1/A
+    elif isinstance(expr, Pow) and expr.args[1] in (-1, -1.0) and len(expr.args) == 2:  # 1/A
```
MichaelClerx: You're going to have to explain this one to me 😆 

If it's about sympy types no longer being considered equal, then
1. How does that affect comparison with a Python float/int? `(-1.0 == -1) = True`, so in the above, if the comparison triggers a cast to float or int, either way we're fine? If not, can we trigger a conversion to a python built-in (float or int) and compare with a literal? This should catch all edge cases then, no? Regardless of new types they introduce
2. I thought at first this wouldn't work, because if expr.args[1] is a sympy type (and if it isn't then the old code should work), then a check based on `is` should fail. However, I checked and the `in` check does try an `==` comparison if all else fails: https://docs.python.org/3/reference/expressions.html#membership-test-operations
Still though. Looks like quite a tricky bit of code. Can we replace with something easier to read / less likely to go wrong?

kwabenantim: Thanks @MichaelClerx. Now we are also get things like this when comparing with Python floats and ints:

```py
>>> Number(1) == 1
True
>>> Number(1.0) == 1.0
True
>>> Number(1) == 1.0
False
>>> Number(1.0) == 1
False
>>> cos(0) == 1
True
>>> cos(0) == 1.0
False
```

This is probably more robust:

```py
expr.args[1].is_number and float(expr.args[1]) == -1.0
```

Since:
```py
>>> cos(x).is_number
False
>>> cos(0).is_number
True
>>> cos(0, evaluate=False).is_number
True
>>> cos(0, evaluate=False) == 1
False
>>> cos(0, evaluate=False) == 1.0
False
>>> float(cos(0, evaluate=False)) == 1.0
True
```

kwabenantim: Fixed in a732fd5

MichaelClerx (approved):

